Issue #21, opened 2018-10-31
Fix production build


wbobeirne:
`yarn build` fails out right now due to TypeScript errors. Get those resolved so we can get a production build out and into the extension store at some point.

PR #28, opened 2018-11-16
Update webpack.config.js

caffeinum:
Addresses #21:

Yes, it won't make a 100% production build, but at least it makes a minified css.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -111,7 +111,7 @@ module.exports = {
           },
           {
             loader: 'ts-loader',
-            options: { transpileOnly: isDev },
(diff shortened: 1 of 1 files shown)


wbobeirne:
While this does fix the build, it only does so by masking the actual errors that are surfaced by this option. Ideally I'd like to fix those up instead. I'll leave this open in case this is the direction it goes to get a build out, but hopefully I can put up a PR that fixes the actual typescript errors.

wbobeirne:
I ended up getting this one solved with #30. Thanks for the contribution though!

PR #30, opened 2018-11-17
Enable production builds

### LINKED ISSUE #21: Fix production build

wbobeirne:
Closes #21.

### What This Does

* Fixes all outstanding Typescript errors that prevented the build from compiling
* Switched from static hardcoded HTML templates to generating them with `html-webpack-plugin` to allow for js-injected CSS in dev and static CSS files in production.
* Updates typescript
* A little bit of house cleaning


wbobeirne:
Very soon. I'd say I'm looking at about a week more of fixes and features before I put it out there.

PR #43, opened 2018-11-26
Reset connection settings

wbobeirne:
Adds a button to the settings menu that resets your node & password. In the future, we'll probably want a more elegant user swap behavior, but this'll do for now.

capture

### DIFF
--- a/src/app/components/Settings/index.tsx
+++ b/src/app/components/Settings/index.tsx
@@ -1,9 +1,10 @@
 import React from 'react';
 import { connect } from 'react-redux';
-import { Form, Select, Checkbox } from 'antd';
+import { withRouter, RouteComponentProps } from 'react-router';
+import { Form, Select, Checkbox, Button, Modal } from 'antd';
 import { SelectValue } from 'antd/lib/select';
 import { CheckboxChangeEvent } from 'antd/lib/checkbox';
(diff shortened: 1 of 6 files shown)


wuestenigel:
Hey, unfortunately I cannot see any settings page as I would like to reset all to begin from scratch with my configuration.

Unfortunately I got stuck here: When I press on Connect, same reappear over and over again. No way to insert Macroon oder any other config. Of course I visited the SSL link in the error message.

Any ideas?

bildschirmfoto 2019-01-16 um 11 07 25

wbobeirne:
If you haven't gotten past this screen, then you haven't had any of your connection settings saved. Simply refreshing will start you over.

It looks like you'll need to set some addiitonal `tlsextraip` in your lnd.conf. I assume you have:
```
tlsextraip=192.168.178.65
```
You can add multiple, so add one below that as
```
tlsextraip=0.0.0.0
```

You'll need to delete your tls certs and re-run to get it to work.

wuestenigel:
Hey, I totally refreshed everything. Deleted the Local Storage of my Chrome Browser, removed the Extension and so on.

I'm wondering there is no possibility to add some macaroon or admin files in Joule.

wbobeirne:
You do that after you get past this screen. You'll need to do the fix I just posted above before you add macaroon files.

PR #54, opened 2018-11-30
Invoice display bug fix

PatrickLemke:
Fixes a bug where only the 30 oldest invoices were displayed. Now the 30 newest invoices will be displayed in the interface.


wbobeirne:
Great catch, I never tested going over 30 invoices. Thanks so much for the contribution ❤️

PR #74, opened 2018-12-09
Fix typo

philippgille:
- Typo in "You have a few options **fow** how you can do that"
- Wrong location for the full stop in "[...] soon.) You [...]"
- Add "for example" in order for the sentence to be correct even when there will be more options in the future


wbobeirne:
Thanks!

wbobeirne:
Ah crap, I just realized this went into `master`. I'll leave as is, but if you have future contributions, please merge into `develop`. `master` reflects the latest release, not the latest development. Not at all your fault though, I haven't setup contributing guidelines!

Issue #45, opened 2018-11-27
Make contribution guidelines, issue templates, code standards


wbobeirne:
In order to promote contributors, we need the following:

* CONTRIBUTORS.md that comes with any information or quirks
* Issue templates for features & bug reports
* Pull request template to standardize what those look like
* A prettier config, and husky pre-commit hook to auto-prettier. Add prettier checks to CI.

PR #80, opened 2018-12-10
Channel progress bars

PatrickLemke:
This pull request adds channel progress bars to the channel overview.

Let me know what you think, I can make changes to the style if needed. This also adds Bootstrap to the Project so if this is not wanted, I can try to accomplish the task in a different way.

There is still one bug in this which I don't understand:
When the plugin is clicked the Channel Balance does not display, only if you maximize the window or e.g. click on settings and then it shows up. If someone knows how to fix it I would be grateful.

### DIFF
--- a/src/app/components/ChannelList/ChannelRow.tsx
+++ b/src/app/components/ChannelList/ChannelRow.tsx
@@ -34,6 +34,9 @@ export default class ChannelRow extends React.Component<Props> {
             {' / '}
             <Unit value={channel.capacity} />
           </div>
+          <div className="ChannelRow-info-progress-bar-background">
+            <div className="ChannelRow-info-progress-bar" style={{width: Math.round(parseFloat(channel.local_balance) / parseFloat(channel.capacity) * 100) + '%'}}></div>
+          </div>
         </div>
       </div>
     );
--- a/src/app/components/ChannelList/ChannelRow.less
+++ b/src/app/components/ChannelList/ChannelRow.less
@@ -39,5 +39,20 @@
       font-size: 0.7rem;
       opacity: 0.8;
     }
+
+    &-progress-bar-background {
+      width: 100%;
+      height: 5px;
+      background-color: #f1f1f1!important;
+      border-radius: 4px;
+      opacity: 0.9;
+      margin-top: 2px;
+    }
+  
+  &-progress-bar {
+      height: 5px;
+      background-color: @primary-color !important;
+      border-radius: 4px;
   }
+}
 }


src/app/components/ChannelList/custom-progress.css
```diff
@@ -0,0 +1,8 @@
+.progress {
+    height: 5px;
+    opacity: 0.7 !important;
+}
+
+.progress-bar {
+    background-color: #7642FF !important;
```
wbobeirne: Should be a LESS variable

src/app/components/ChannelList/ChannelRow.tsx
```diff
@@ -4,6 +4,8 @@ import Identicon from 'components/Identicon';
 import Unit from 'components/Unit';
 import { ChannelWithNode } from 'modules/channels/types';
 import './ChannelRow.less';
+import {Progress} from 'reactstrap';
+import './custom-progress.css';
```
wbobeirne: All styles in src should use LESS rather than CSS.

package.json
```diff
@@ -49,11 +50,11 @@
     "babel-polyfill": "6.26.0",
     "bip39": "2.5.0",
     "bn.js": "4.11.8",
+    "bootstrap": "^4.1.3",
     "classnames": "^2.2.6",
     "clean-webpack-plugin": "^0.1.19",
     "copy-webpack-plugin": "4.6.0",
     "crypto-js": "3.1.9-1",
-    "css-loader": "^1.0.0",
```
wbobeirne: Why was this removed? It's still being used in the webpack config.

src/app/components/ChannelList/custom-progress.css
PatrickLemke: fixed

src/app/components/ChannelList/ChannelRow.tsx
PatrickLemke: fixed

package.json
PatrickLemke: Dump mistake on my part, fixed.

PatrickLemke:
Thanks for giving such detailed feedback, in hindsight this was a very embarrassing PR. I didn't follow any of your guidelines and added unnecessary code, so shame on me.
Thanks for being so nice, hopefully those changes make more sense now. Also let me know if want any other changes to the design, etc.

wbobeirne:
No worries man! Appreciate you taking the feedback, I'll take another look later. I'm also working on contributing guidelines (#45) which should make this kind of stuff more explicit.

wbobeirne:
I made a couple of changes that I've PR'd against your PR over here: https://github.com/PatrickLemke/joule-extension/pull/2

Hope you don't mind! I thought that would be easier than telling you some of my concerns. Let me know if you have any thoughts.

PR #92, opened 2018-12-15
BTCPay Server onboarding

### LINKED ISSUE #79: Add BTCPay Server option to onboarding
Allow users to quickly connect with the BTC pay server by providing the config string from `/server/services/lnd/BTC` and having that set them up. Make it a separate option from "local node" and "remote node" so that instructions can be laid out for them.

wbobeirne:
Closes #79. Will want a review from @NicolasDorier on this one.

### Description

Adds an option to the node selector for BTCPay Server. Uses the BTCPay Server config format for filling out URI & macaroons.

### Steps to Test

1. Setup a BTCPay Server
2. Go through Joule onboarding, select BTCPay Server
3. Follow the instructions and fill out the config data
4. Confirm you connect correctly

### Screenshots

onboarding

^Click to embiggen

### DIFF
--- a/src/app/components/SelectNode/index.tsx
+++ b/src/app/components/SelectNode/index.tsx
@@ -163,7 +163,7 @@ class SelectNode extends React.Component<Props, State> {
   private handleBTCPayServerConfig = (config: BTCPayServerConfig) => {
     const macaroons = {
       adminMacaroon: config.macaroon,
-      readonlyMacaroon: config.restrictedMacaroon || config.macaroon,
+      readonlyMacaroon: config.readonlyMacaroon || config.macaroon,
     };
     this.setState(macaroons, () => {
       this.props.checkAuth(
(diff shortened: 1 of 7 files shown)


NicolasDorier:
This is awesome: Though can you do this:

In the configuration file, right now you can find the `macaroon`.
In a future version, I will add the `readonlymacaroon`.

Can you, in your code consider that if `readonlymacaroon` is null or does not exist, then use `macaroon` for both?

That way, when I update BTCPay it will properly use the right one.

src/app/components/SelectNode/index.tsx
```diff
@@ -148,6 +160,20 @@ class SelectNode extends React.Component<Props, State> {
     }
   };
 
+  private handleBTCPayServerConfig = (config: BTCPayServerConfig) => {
+    const macaroons = {
+      adminMacaroon: config.macaroon,
+      readonlyMacaroon: config.restrictedMacaroon || config.macaroon,
```
NicolasDorier: use `readonlyMacaroon` instead of `restrictedMacaroon`

NicolasDorier:
Ah actually you do this, just rename `restrictedMacaroon` to `readonlyMacaroon`.

wbobeirne:
Thanks for that, the change has been made. I'll be merging shortly.

PR #108, opened 2018-12-22
Add context menu item for selected bolt11 payment requests

### LINKED ISSUE #9: Context menu entry for highlighted payments & QR codes
If you right click on a highlighted BOLT-11 payment string / qr code, it would be awesome to have a one-click option to prompt you to pay, rather than having to copy / paste it in.

jamaljsr:
Closes #9 

### Description

Added a context menu item  to "Pay Lightning Invoice" when a BOLT-11 payment request string is highlighted on the page and right-clicked. When this menu item is clicked, the payment popup is displayed allowing you to confirm the payment.

### Steps to Test

1. Visit a website that accepts lightning payments 
2. Click Buy button in order to be presented with a Lightning payment request string
3. Highlight the string then right-click
4. Click on the `Pay Lightning Invoice` menu item
5. Click `Confirm` in the popup to send the bitcoin

### Screenshots

joule_context_menu

### DIFF
--- a/src/background_script/index.ts
+++ b/src/background_script/index.ts
@@ -1,6 +1,7 @@
 import qs from 'query-string';
-import { browser, Runtime } from 'webextension-polyfill-ts';
-import { OriginData } from 'utils/prompt';
+import { browser, Runtime, Menus } from 'webextension-polyfill-ts';
+import { getOriginData, OriginData } from 'utils/prompt';
+import { isValidPaymentReq } from 'utils/validators';
 import { PROMPT_TYPE } from '../webln/types';
 import getNodeInfo from './getNodeInfo';
 
@@ -53,13 +54,11 @@ function openPrompt(request: PromptRequest): Promise<any> {
 
 // Background manages communication between page and its windows
 browser.runtime.onMessage.addListener((request: any) => {
-  if (request && request.application === 'Joule' && request.prompt) {
-    // Special case -- get info requires no prompt, just respond
-    if (request.type === PROMPT_TYPE.INFO) {
-      return getNodeInfo().then(data => ({ data }));
-    }
+  // abort early for other extensions
+  if (!request || request.application !== 'Joule') return;
 
-    // WebLNProvider request, will require window open
+  // WebLNProvider request, will require window open
+  if (request.prompt) {
     return openPrompt(request)
       .then(data => {
         return { data };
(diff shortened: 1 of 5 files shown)


wbobeirne:
This is a really clever work-around for the extension API's lack of a way to show conditionally based on the content highlighted. I had hoped it'd work even without highlighting the BOLT-11 string, just by reading the content of the element, but it doesn't look like there's a straightforward way to do that. I'm going to dig in a little bit and see if that can be done, but if not, I'll merge this in.

Thanks for the contribution!

jamaljsr:
I had not even consodered the case where the user right-clicks without highlighting first. I think it is should be possible if the mousedown event captures the associated DOM element. Then it’s just a matter of navigating up to the parent tag and getting the innerText. I will try this tomorrow and let you know how i make out.

wbobeirne:
That sounds great, thanks for looking into it.

src/background_script/index.ts
```diff
@@ -51,15 +52,20 @@ function openPrompt(request: PromptRequest): Promise<any> {
   });
 }
 
+// Need to store the payment request string in a background variable
+// to support the case where the user right-clicks on text without
+// selecting it first. In this situation, the background script does
+// not have access to the page dom to pull the string in the context
+// menu onclick handler.
+let currentPaymentRequest: string;
```
jamaljsr: I personally am not a fan of this line. I am not sure if global variables in background scripts are considered a code smell, but I tried searching for a better solution and couldn't come up with anything better. I'm very open to suggestions here if you have any.

wbobeirne: Nope, this makes sense to me. However, it might be good to refactor this section out into its own file. I've got another feature on another branch that does that so that background script isn't one big blob, I'll probably merge this as-is after I do some more testing and then refactor in a later commit. I appreciate you looking out for code quality though!

jamaljsr: Good point about the refactor. I agree 100%. I'll let you handle that and move on to another issue.

Thanks for the feedback.

src/background_script/index.ts
```diff
@@ -68,4 +74,36 @@ browser.runtime.onMessage.addListener((request: any) => {
         return { error: err.message };
       });
   }
+
+  // Special cases
+  switch (request.type) {
```
wbobeirne: This switch needs to come before the `openPrompt` call above, otherwise `PROMPT_TYPE.INFO` opens a prompt. You can test this out by calling
```
webln.enable().then(() => webln.getInfo())
```
in your console.

jamaljsr: Oh wow, good catch. 

In my head, while refactoring this function, I intended to make the sender of `PROMPT_TYPE.INFO` messages set `prompt: false` on the message, since it shouldn't actually display a prompt. Clearly this slipped my mind.

I just looked into the usages of `PROMPT_TYPE.INFO` and realized that the sender of this message is in webln and would require a bit of refactoring to set `prompt: false`. So I'll make the update you suggested by putting the `switch` before the `openPrompt`.

wbobeirne: Great, thanks! There will be additional cases that could or would be prompts, but won't prompt the user (e.g. invoices that have a fixed amount can be auto-generated, payments that you have an allowance setup for can automatically be done) so there will still be things that are `prompt: true` that don't _actually_ prompt the user. I totally get where you were coming from though, a lot of this message passing stuff isn't very clear.

jamaljsr: Ok, I get it now. That makes sense then.

jamaljsr: I made the change requested above.

PR #117, opened 2018-12-28
Ask for password only once

### LINKED ISSUE #96: Make required password to spend optional.
Make required password to spend optional.

wbobeirne:
Closes #96 

### Description

Adds a little background script to cache the password after you've entered it. Before asking for password, attempt to fetch from that cache. This method should be completely safe for a few reasons:

1. The password is stored in memory, not using any long-term storage. Closing the browser will remove the password from memory, and any updates to the extension will clear the background context, so it won't ever be immediately available without the user entering it.
2. The password is sent around using [`runtime.sendMessage()`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/sendMessage), which only communicates to contexts created by the extension, and doesn't communicate to the content script.
3. All payments and invoice generation still require a prompt.
4. You can uncheck the "Don't ask me again" checkbox if you want and no messages will be sent anywhere.

### Steps to Test

1. Perform an action that requires a password (e.g. click Deposit). Leave the checkbox checked.
2. Close the popup, and perform that action again. Confirm you weren't asked for the password.
3. Open the menu dropdown and click "Lock account". Perform the action again and confirm you were asked for the password.
4. Enter the password, but uncheck the checkbox. Close the popup, perform the action again, and confirm you were asked for the password again.

### Screenshots

screen shot 2018-12-28 at 1 59 16 pm

screen shot 2018-12-28 at 1 59 06 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,18 @@ to the joule folder and run `yarn link webln`.
 
 NOTE: After making changes, you'll need to close and re-open the extension to load the latest build.
 
+Redux DevTools:
+1. Open the extension popup or full page
+2. Right click on the background
+3. Choose Redux Devtools -> Open Remote DevTools
+4. A new window will open displaying the Redux actions list
+
+React DevTools:
+1. Run `npm install -g react-devtools`
+2. Be sure to use `yarn run dev` to build the app
+2. Run `react-devtools` in a new Terminal
+3. A new window will open displaying the React vdom inspector
+
 ## Building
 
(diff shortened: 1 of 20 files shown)


jamaljsr:
Great addition! I tested it and it works as described.

wbobeirne:
Awesome, thanks for giving it a look. I'll merge after I've tested and merged #108, since I know there will be conflicts.

PR #127, opened 2019-01-02
Fixed deposit button not working if password is dismissed

### LINKED ISSUE #105: Deposit Button not working if password is dismissed
### Description of the Issue

When the Deposit Button is clicked and the password is not entered, but the 'x' button is clicked, the deposit button can not be clicked again.

### Description of Your Setup

*Browser & version*: Chrome Version 71.0.3578.98 (64 bit)
*Operating system*:Mac OS Mojave 10.14.2
*Node software & version*: LND version 0.5.1

Video of Issue

[Issue Deposit.mp4.zip](https://github.com/wbobeirne/joule-extension/files/2702559/Issue.Deposit.mp4.zip)

callmekurisu:
Closes #105 

### Fixed deposit button not working if password dismissed

Initially set deposit modal to true on local state. Then used a modal state from the reducer until the user successfully enters the password. Cancel password action sets modal back to false since before it was stuck on true after CANCEL_PASSWORD action. After entering password the deposit button will still show address until closing the app. Only thing is that subsequent attempts after dismissing password prompt requires double tap instead of single tap.

*edit - solved double tap issue.


callmekurisu:
Ah shoots. Just noticed the "lock account" feature. This is causing a problem with deposit modal. Will try and fix it soon.

wbobeirne:
Much simpler fix is up at #130, so I'm going to close this out. Thanks for taking a crack at it though!

PR #135, opened 2019-01-11
Add support for Litecoin

### LINKED ISSUE #132: Support non-BTC nodes (Litecoin)

Since Bitcoin isn't the only currency Lightning can handle, Joule should play nice with any and all currencies. Right now Litecoin is the only other one I'm aware of, but it should be supported in a way that adding additional currencies is quite easy.

majestic84:
Closes #132

### Description

This PR adds support for Litecoin. Joule will be able to detect the type of node it is connected to by the "chains" attribute of the getinfo output.

### Steps to Test

1. Have a fully synced litecoind instance running and configured
2. Have a Lightning Node running and connected to litecoind
3. Run Joule and connect to the node

### Screenshots
https://imgur.com/a/H5CDDnx
https://imgur.com/a/gc7RrBU
https://imgur.com/a/V4EQFle
https://imgur.com/a/CMjvcqr


src/app/components/AccountInfo/index.tsx
```diff
@@ -71,7 +78,7 @@ class AccountInfo extends React.Component<Props, State> {
               </div>
               <div className="AccountInfo-top-info-balances">
                 <span>Channels: <Unit value={account.channelBalance} /></span>
-                <span>Bitcoin: <Unit value={account.blockchainBalance} /></span>
+                <span>{blockchain} <Unit value={account.blockchainBalance} /></span>
```
coblee: You missed a colon.

```suggestion
                <span>{blockchain}: <Unit value={account.blockchainBalance} /></span>
```

src/app/components/InvoiceForm/index.tsx
```diff
@@ -241,21 +247,26 @@ class InvoiceForm extends React.Component<Props, State> {
   };
 
   private updateBothValues = (name: string, val: string) => {
-    const { fiat, rates } = this.props;
+    const { fiat, rates, nodeInfo } = this.props;
     const { denomination } = this.state;
     let { value, valueFiat } = this.state;
+
+    let rateSelector = 'btcRate';
```
coblee: it will be simpler to just key off of the chain name 'bitcoin' and 'litecoin' instead of using a new key 'btcRate' and 'ltcRate'

src/app/components/AccountInfo/DepositModal.tsx
```diff
@@ -41,9 +42,15 @@ class DepositModal extends React.Component<Props> {
       isOpen,
       onClose,
       hasPassword,
+      nodeInfo,
     } = this.props;
     const isVisible = !!isOpen && !!(hasPassword || fetchDepositAddressError);
 
+    let blockchain = "Bitcoin";
+    if (nodeInfo && nodeInfo.chains[0] === 'litecoin') {
```
coblee: will nodeInfo ever not exist?

coblee: if it's always there, then just set the name to nodeInfo.chains[0] capitalized.

majestic84: I thought about doing this, but IntelliJ warns that nodeInfo could possible be null.

coblee:
Ideally everything is key’d off of `nodeInfo.chains[0]`. So you just look it up the name/symbol from a constant dict. Even the denomination symbols and names should be a constant lookup like `symbos[nodeInfo.chains[0]][denomination]`. So no need to litter the code with if/else statements.

Also, maybe don’t pass `nodeInfo` around. Just pass `chain` to the methods that need it.

majestic84:
Thanks @coblee. @wbobeirne please don’t merge until I undress these improvements.

wbobeirne:
Hey @coblee, thanks for taking a look at this. He's right on with some of the feedback, but I've got some notes here myself. If you can hold off on making changes for a few hours, I'll have a thorough review later tonight when I can get back to my computer.

src/app/modules/rates/sagas.ts
```diff
@@ -8,7 +8,9 @@ import types from './types';
 export function* handleFetchRates(): SagaIterator {
   try {
     const fiats = Object.keys(Fiat);
-    const rates = yield call(apiFetchRates, 'BTC', fiats);
+    const btcRate = yield call(apiFetchRates, 'BTC', fiats);
+    const ltcRate = yield call(apiFetchRates, 'LTC', fiats);
```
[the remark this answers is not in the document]
coblee: I suggest we add an array constant for supported chains which has just 'bitcoin' and 'litecoin' in it. fetch rates should then iterate through the support chains, lookup the symbol for that chain ('BTC', 'LTC', etc) and fetch the rates for that chain. So adding a new coin shouldn't have to change this method.

majestic84:
Thanks fellas. I’ll get started with these improvements tomorrow!

src/app/components/AccountInfo/DepositModal.tsx
```diff
@@ -7,20 +7,22 @@ import Copy from 'components/Copy';
...
   hasPassword: boolean;
-  nodeInfo: AppState['node']['nodeInfo'];
 }
 
 interface DispatchProps {
   getDepositAddress: typeof getDepositAddress;
 }
 
 interface OwnProps {
+  chain: CHAIN_TYPE;
```
wbobeirne: This is actually a `StateProps` prop, since it comes from the redux store. It's also usually good to type selectors as `ReturnType<typeof selectorFnName>;` to make refactors easier, if you change the selector.

src/app/components/InvoiceForm/index.tsx
```diff
@@ -28,6 +28,7 @@ interface DispatchProps {
 }
 
 interface OwnProps {
+  chain: CHAIN_TYPE;
```
wbobeirne: Also StateProps

src/app/components/SendForm/index.tsx
```diff
@@ -6,16 +6,15 @@ import BigMessage from 'components/BigMessage';
...
 import { connect } from 'react-redux';
-
-interface StateProps {
-  nodeInfo: AppState['node']['nodeInfo'];
-}
+import { blockchainDisplayName, CHAIN_TYPE } from 'utils/constants';
+import { getNodeChain } from 'modules/node/selectors';
 
 interface OwnProps {
+  chain: CHAIN_TYPE;
```
wbobeirne: Also StateProps

src/app/components/Settings/index.tsx
```diff
@@ -18,19 +18,22 @@ import {
...
+import { getNodeChain } from 'modules/node/selectors';
 
 type SettingsKey = keyof AppState['settings'];
 
 interface StateProps {
   settings: AppState['settings'];
-  nodeInfo: AppState['node']['nodeInfo'];
+}
+interface OwnProps {
+  chain: CHAIN_TYPE;
```
wbobeirne: Also StateProps

src/app/components/TransactionList/TransactionRow.tsx
```diff
@@ -25,19 +21,15 @@ interface OwnProps {
   status: 'complete' | 'pending' | 'rejected' | 'expired';
   pubkey?: string;
   delta?: BN | false | null;
+  chain: CHAIN_TYPE;
```
wbobeirne: Also StateProps

src/app/components/Unit.tsx
```diff
@@ -4,26 +4,27 @@ import {
...
   isNoFiat: AppState['settings']['isNoFiat'];
-  nodeInfo: AppState['node']['nodeInfo'];
 }
 
 interface OwnProps {
   value: string;
   hideUnit?: boolean;
   showFiat?: boolean;
   showPlus?: boolean;
+  chain: CHAIN_TYPE;
```
wbobeirne: Also StateProps

src/app/modules/rates/sagas.ts
```diff
@@ -8,9 +8,9 @@ import types from './types';
 export function* handleFetchRates(): SagaIterator {
   try {
     const fiats = Object.keys(Fiat);
-    const btcRate = yield call(apiFetchRates, 'BTC', fiats);
-    const ltcRate = yield call(apiFetchRates, 'LTC', fiats);
-    const rates = { btcRate, ltcRate }
+    const bitcoin = yield call(apiFetchRates, 'BTC', fiats);
+    const litecoin = yield call(apiFetchRates, 'LTC', fiats);
+    const rates = { bitcoin, litecoin };
```
wbobeirne: It would be great to make this programmatic so that new CHAIN_TYPEs have their rates fetched, but I'd be happy to make that change myself.

src/app/prompts/invoice.tsx
```diff
@@ -21,14 +22,17 @@ interface StateProps {
   fiat: AppState['settings']['fiat'];
   isNoFiat: AppState['settings']['isNoFiat'];
   rates: AppState['rates']['rates'];
-  nodeInfo: AppState['node']['nodeInfo'];
+}
+
+interface OwnProps {
+  chain: CHAIN_TYPE;
```
wbobeirne: Also StateProps

src/app/prompts/payment.tsx
```diff
@@ -26,12 +27,16 @@ interface StateProps {
   denomination: AppState['settings']['denomination'];
 }
 
+interface OwnProps {
+  chain: CHAIN_TYPE;
```
wbobeirne: Also StateProps

src/app/utils/constants.ts
```diff
@@ -8,39 +13,58 @@ export const LND_DIR = {
   LINUX: '~/.lnd/data/chain/bitcoin/*',
 };
 
+export enum CHAIN_TYPE {
+  BITCOIN = 'bitcoin',
+  LITECOIN = 'litecoin',
+}
+
+export const blockchainLogos: { [key in CHAIN_TYPE]: React.ComponentType<CustomIconComponentProps> } = {
```
wbobeirne: These actually aren't antd icons, they're React SVG components (As handled by `@svgr/webpack`.) A correct and simpler way to type this would be simply `React.ReactNode`.

src/app/components/AccountInfo/index.tsx
```diff
@@ -74,11 +71,11 @@ class AccountInfo extends React.Component<Props, State> {
             <div className="AccountInfo-top-info">
               <div className="AccountInfo-top-info-alias">{account.alias}</div>
               <div className="AccountInfo-top-info-balance">
-                <Unit value={account.totalBalance} showFiat />
+                <Unit value={account.totalBalance} chain={chain} showFiat />
```
wbobeirne: `<Unit>` is already a connected component, and should be running the selector itself, instead of adding it to every instance of it.

src/app/modules/rates/sagas.ts
majestic84: Ah yes, I forgot about this suggestion. Will do!

src/app/utils/constants.ts
majestic84: Thanks for the quick feedback! I’ll have this done tonight :)

src/app/components/AccountInfo/DepositModal.tsx
jamaljsr: It looks like the `getNodeChain` function return type is `null || string` but you want it to be `CHAIN_TYPE`. You can fix this by updating the selector to be:
```
export const getNodeChain = (s: AppState) => ((s.node.nodeInfo && s.node.nodeInfo.chains[0]) || 'bitcoin') as CHAIN_TYPE;
```

Would it make more sense to lookup the values of `blockchainDisplayName`, `blockchainLogos`, `denominationSymbols` and `denominationNames` immediately after the `nodeInfo` is fetched from the API, doing it only once. You could do these lookups in the `handleGetNodeInfo` saga and pass it along with the `GET_NODE_INFO_SUCCESS` action, then store these values in the node redux state. 

Now in the UI components, you are just pulling the same string values from state instead of doing the lookups every time the components are rendered. And you no longer need to import `CHAIN_TYPE` and `getNodeChain` into many components to do the lookups.

wbobeirne: Yep, @jamaljsr has it. Typescript is seeing `'bitcoin'` here and thinking that it could be just any ol' string. The type might have been correctly inferred by changing `nodeInfo`'s type to be an array of CHAIN_TYPEs instead of strings, and to use `CHAIN_TYPE.BITCOIN` instead of `'bitcoin'`.

As for the redux store stuff, I'm fine with it as-is since I like keeping redux state light, but if there end up being a ton of UI changes as a result of chain type, it might be wise to bundle them all together into something like what @jamaljsr described.

jamaljsr: @wbobeirne has a good point about keeping the redux state light. An alternative approach, which is a middle ground, is to change your selector to return an object with the lookup values instead of just the chain. It still reduces the duplication of lookup code in your components, but doesn't pollute the redux state. 

Example:
```
export interface ChainInfo {
  displayName: string;
  denominationSymbols: DenominationMap;
  denominationNames: DenominationMap;
  blockchainLogo: React.ComponentType<CustomIconComponentProps>;
}

export const getChainInfo = (s: AppState): ChainInfo => {
  const chain = ((s.node.nodeInfo && s.node.nodeInfo.chains[0]) || 'bitcoin') as CHAIN_TYPE;
  return {
    displayName: blockchainDisplayName[chain],
    denominationSymbols: denominationSymbols[chain],
    denominationNames: denominationNames[chain],
    blockchainLogo: blockchainLogos[chain],
  };
}
```
Component:
```
import { ChainInfo, getChainInfo } from 'modules/node/selectors';
...
interface StateProps {
  chainInfo: ChainInfo
}
...
<Modal
  title={chainInfo.displayName + " Address"}
>

<Icon component={chainInfo.blockchainLogo} />
...
export default connect<StateProps, DispatchProps, OwnProps, AppState>(
  state => ({
    chainInfo: getChainInfo(state),
  }),
``` 
Just throwing this out there as food for the conversation, I don't want to over-complicate it. 😄

majestic84:
Doh I went ahead and moved everything into the state as @jamaljsr had suggested. It's late here now. Tomorrow I'll revert it back to keep the redux state light as you said and finish the cryptocompare bit,

majestic84:
Hi @wbobeirne 

This latest commit addresses all of your recommendations except for the suggested changes to app/lib/cryptocompare.ts and app/modules/rates/sagas.ts (handleFetchRates). I started writing apiFetchRates to accept an array of coins from Object.keys(CHAIN_TYPE) but then when I started getting into it I felt like i was over complicating the whole process, so I figured I'd go ahead and commit everything else for review first.

Please feel free to let me know if you'd like to me to change anything else, and thanks again for your feedback.

majestic84:
I just committed a proposed solution to fetching the rates from an array of coins, as recommended by one of your earlier comments. Take a look at that and let me know what you think :)

src/app/components/Unit.tsx
```diff
@@ -34,6 +40,7 @@ class Unit extends React.Component<Props> {
       denomination,
       isFiatPrimary,
       isNoFiat,
+      chain
```
coblee: I believe the the style is to keep a comma at the end of the last element.

```suggestion
      chain,
```

src/app/components/AccountInfo/DepositModal.tsx
```diff
@@ -86,7 +90,7 @@ class DepositModal extends React.Component<Props> {
 
     return (
       <Modal
-        title="BTC Address"
+        title={blockchainDisplayName[chain] + " Address"}
```
coblee: Should this be `coinSymbols[chain]` instead to keep it being "BTC Address"? Unless you want it to display "Bitcoin Address" now.

majestic84: Good call, Charlie. Pushing this up right now.

src/app/utils/constants.ts
majestic84: Typing blockchainLogos as React.ReactNode causes Travis CI to fail. I've set it back to React.ComponentType<CustomIconComponentProps> for now, as that was working well.

Issue #71, opened 2018-12-06
Show pending channels


wbobeirne:
When you create a new channel, it's visible via https://api.lightning.community/rest/index.html#v1-channels-pending until it's confirmed. We should query and display this as well.

jnewbery:
Sounds good.

There's also the rather user-unfriendly behaviour that `walletbalance` does not include balances from pending channels. If launched with autopilot, the user's balance decreases when channels are opened until they reach the required number of confirmations. Perhaps it'd be friendlier behaviour if the balance shown in Joule also showed balances from pending channels?

wbobeirne:
Yes, I had exactly this experience today. I'll probably just change that display to use `totalBalancePending` rather than `totalBalance`. I had questioned whether it was a worse experience to show a user's balance they couldn't access yet, or to not show it at all, I definitely think I chose wrong. I can probably add a little icon / tooltip when some is pending that, when hovered over, shows your available balance.

bitcoinuser:
Hi,

In my opinion you should show total balance (including the one that couldn't be acessed yet). 

I have an money app that it's not related to crypto but have an interesting structure.

1- They show on main window the (Total Balance = Money in account)
2- When the user click on Total Balance, appear a new window that shows:

* **Money in account** (total balance): 50 dollars
* **Available money** (Money that you can spend now): 40 dollars
* **Pending deposit** (money that you own but need some time to be able to be spent): 10 dollars
* **Money to be released** (Money the user redeemed to an external bitcoin address): 40 dollars

3- In front of **Available money** could have a button (withdraw), so when users click on it, joule will ask the total amount to withdraw and the bitcoin address to send it. When user fill the information, joule will tell user that channels a, b or c will be closed to proceed the withdraw and that the process will take some time. 

4- When user click on **Money to be released** could show the information of when approximately the withdraw process will end. The same thing for **Pending deposit**.

5- I think joule could have only the option to deposit bitcoins to open channels. Maybe an option to user select if he wants to open channels automatically or no. But I think the process to send on-chain bitcoin should appear only on withdraw specified on item 3.

6- On main window of Joule you should keep just * **Money in account** and not the Channels and Bitcoin as it's now.

Thanks.

callmekurisu:
I got pending force close on the list. Should be able to get the pending open channels done soon. 

![image](https://user-images.githubusercontent.com/13033037/50546901-7f2c1800-0bf5-11e9-9b80-8bab2f67584d.png)

Issue #38, opened 2018-11-19
Open & close channels


wbobeirne:
We have a list of them, so it would be nice to be able to open new channels and close existing ones. This would also be a good opportunity to list recommended channels to connect to(?).

Potentially blocked on https://github.com/lightningnetwork/lnd/issues/2197

PatrickLemke:
Thread in lnd says issue is fixed so this feature could be implemented.

PR #136, opened 2019-01-12
Open new channels

### LINKED ISSUE #118: Create common-use component for amount entry
Entering an amount, with a denomination selector, optionally showing fiat or providing a fiat input are used a lot. I've been re-implementing it all over the place, so we should probably have a reusable component for this. Ideally it's flexible enough to fit all of the above use-cases.

wbobeirne:
Completes half of #38 & closes #118.

### Description

Adds the ability to open new channels. You specify a connection string, a funding amount, and optionally a push amount and if it should be private.

This also adds an `<AmountField>` component for amount entries, and is used on InvoiceForm. I didn't change it everywhere yet due to keep this small, but I'll follow up with another PR to change them all.

### Steps to Test

1. Click the top right menu and select "Open channel", confirm it opens the modal.
2. Attempt to submit the form, confirm it pops up an error.
2. Attempt to enter an invalid connection string, confirm it highlights the field in red.
3. Enter a valid connection string, confirm it accepts it.
4. Attempt to enter an amount that exceeds your available on-chain balance, confirm it shows an error.
5. Attempt to enter a negative amount, confirm it shows an error.
6. Enter a valid amount, confirm it accepts it.
7. Submit the form, confirm it works (Assuming there are no legitimate errors)
8. Click on "View transaction", confirm you're taken to an explorer with a valid transaction
9. Attempt to submit the same form as before, confirm you get an error about maximum channels

Also make sure the invoice form works still, as its amount field was replaced with the AmountField component created for this.

Note that you won't see the channel in your channel list, since pending channels aren't yet shown (#71) I also didn't include testing instructions for the advanced fields, but my limited testing showed they worked, and you're welcome to give it a shot yourself.

### Screenshots

screens


jamaljsr:
I tested and everything looks great. Very simple flow for the user.

I did run into trouble when I tried to open a channel with the same node that already has a pending channel opening. I clicked on the `Open another` button and entered the same pubkey@host. The success message is displayed but the `Number of pending channels exceed maximum` message is returned from the lnd API, as displayed in the devtools console. Also, the `View transaction` button links to the same txid as the previous channel opening.

image

If I reload the options page before trying to open the second channel, then the maximum channels error shows correctly in the UI.

wbobeirne:
Awesome, thanks for testing. Should be easy to block subsequent within the same session by looking up if the entered connection string already has a successful txid associated with it.

PR #161, opened 2019-02-10
Implement sending on-chain transactions

### LINKED ISSUE #31: Send bitcoin transactions
There's a form for sending lightning transactions, but bitcoin transactions are also possible using the [`POST /v1/transactions` API](https://api.lightning.community/rest/index.html#v1-transactions).

jamaljsr:
Closes #31 

### Description

This PR adds initial support to make on-chain payments. 

_Note: I'm opening this PR for early feedback. Additional functionality I want to implement is listed below_

### Steps to Test

1. Click the `Send` button on the home screen
2. Click on the `Bitcoin`/`Litecoin` tab
3. Enter the amount and the recipient address
4. Click the `Send` button

### Additional Functionality to Implement

- [x] calculate and display fees (via [earn api](https://bitcoinfees.earn.com/api)?)
- [ ] display the balance locked in channels and allow user to close channels to free up funds
- [x] warn user if the tx will create a small dust utxo (LND will return an error in this case)


src/app/modules/payment/selectors.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  const { fastestFee } = fees;
+  let { halfHourFee, hourFee } = fees;
+
+  if (halfHourFee === fastestFee) {
+    halfHourFee--;
+  }
+  if (hourFee >= halfHourFee) {
+    hourFee--;
+  }
```
wbobeirne: Great catches on this edge case, but another edge case this introduces is if the network is at a very low fee state, the remote possibility that these become `0` could introduce problems. Might be better to _increment_ the other fees instead, for safety's sake.

Also, the `hourFee >= halfHourFee` logic means that if hourFee were 3, and halfHourFee were 2, hourFee would become 2, making them the same. It should probably be `hourFee = halfHourFee - 1`.

src/app/lib/earn.ts
```diff
@@ -0,0 +1,6 @@
+const API_URL = 'https://bitcoinfees.earn.com/api/v1/fees/recommended';
```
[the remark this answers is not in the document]
jamaljsr: To make this decision, I searched high and low trying to find ways to estimate fees. Without direct access to the mempool, we cannot do it ourselves. I found that [Bitgo](https://www.bitgo.com/api/v2/#estimate-transaction-fees) has an API but it requires authentication. Earn.com provides estimates anonymously and they are backed by Coinbase so I feel confident that it'll remain online for the near future. 

LND doesn't provide any details on fee estimation until after the transaction is sent. They use the underlying btcd/bitcoind RPC to get the fee. I have a thought to submit a PR to add a new RPC endpoint to calculate the fee of a transaction without sending it. This way, we can display the exact number of satoshis to the user and we wouldn't need to use the earn API at all. I know this would likely take a long time to get implemented, approved, merged, released, then installed by users. So I figured to go with the earn API for now.

I also looked at a few other lightning wallets to see what they use:
1. Lightning App - no fee estimation, always displays 0 fee
2. Zap - uses earn.com ([ref](https://github.com/LN-Zap/zap-desktop/blob/master/app/lib/utils/api.js#L42))
3. BlueWallet - uses earn.com ([ref](https://github.com/BlueWallet/BlueWallet/blob/master/models/networkTransactionFees.js#L17))

There is still a limitation using this API. The earn.com fee estimates are returned as sats/byte based on the current mempool. Since Joule cannot know how many bytes the tx will be ahead of time, we cannot display the exact number of satoshis the transaction will cost. So right now I just display the sats/byte. It's not ideal, but I feel like it's better than not displaying a fee at all.

src/app/modules/payment/selectors.ts
jamaljsr: I was thinking to default to the fastestFee and work backwards, but I didn't consider the case of a `0` fee. Good point. I'll fix it up.

jamaljsr: I just pushed what I think is a better solution to the original problem. With this selector, I was attempting to fix a UI issue. Multiple RadioButtons were being selected when the fee values from the API were the same for more than one key in the response object. For example: `{ fastestFee: 25, halfHourFee: 25, hourFee: 16 }`. As you saw, my initial solution was to decrement the values slightly to prevent multiple RadioButtons being selected. This in-turn presented other issues with negative numbers like you mentioned. After playing with the adjustments further, I felt like I was adding too much complexity to solve a simple UI problem. I also didn't really like increasing the fee values just to fix the UI. This is someone's money we're spending. Even though it's just a few sats, it didn't sit well in my gut.

I took a step back and looked at the original issue to try to find an alternate solution. I think I came up with something simpler. Instead of using the numeric fees for the RadioButton values (which can contain duplicates), I am now using the response object's key as the value for the RadioButton (ex: `fastestFee` instead of `25`). This solves the multiple button selection UI issue without needing to alter the numeric values. It also eliminates the need for the selector code.

I feel much better about this approach. Let me know what you think.

src/app/lib/earn.ts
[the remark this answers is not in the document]
wbobeirne: Thorough analysis, thanks. I think you made the best decision with that information 👍

src/app/modules/payment/selectors.ts
wbobeirne: Sounds like a great chop of the Gordian knot. Nice work!

jamaljsr:
Ok, I think this is ready for a full review. 

The current dust detection functionality is pretty basic. It's ok but not great IMO. I just found this [coinselect](https://github.com/bitcoinjs/coinselect) library that could be another option for estimating fees more accurately. We just need to give it a list of utxos and the fee rate from earn.com. It will return the exact fee in sats. I want to play with it to see how accurate it is. If it works i'll create a separate PR because this one is already growing pretty big.

For the feature to show the wallet balance locked in channels, I thought of an idea that I think would look nice. Show a modal with a pie chart type graph with 3 slices for where funds are allocated (in channels, on-chain, unavailable/pending). I think I can use the circular progress from antd so we don't need to add a heavy charting library as a dependency. Below the pie chart, show tabs & tables listing further details of how the funds are broken up per channel, utxo & pending txn. This same modal can be linked from the pending icon currently on the home screen as well. I know this may be a bit more code, so I want to do that in a separate PR as well.

src/app/components/SendForm/ChainSend.tsx
```diff
@@ -0,0 +1,272 @@
...
+      );
+    }
+
+    const { amount, isSendAll, address, fee, feeMethod, showMoreInfo } = this.state;
+    const blockchainBalance = account ? account.blockchainBalance : '';
+    const disabled = (!amount && !isSendAll) || !address ||
+      (!!blockchainBalance && !!amount && (new BN(blockchainBalance).lt(new BN(amount))));
+    const dustWarning = isPossibleDust(amount, address, fee) ? (
+      <><Icon type="exclamation-circle" /> Dust warning: This amount may not be spendable</>
+    ) : undefined;
```
wbobeirne: I could never actually get this to show up, no matter how small I made the remainder. Not sure if there's something I was doing wrong here though.

jamaljsr:
Thanks for the review and feedback. I’ll get these items fixed up.

src/app/components/SendForm/ChainSend.tsx
```diff
@@ -0,0 +1,272 @@
...
+          onChangeAmount={this.handleChangeAmount}
+          minimumSats="1"
+          maximumSats={blockchainBalance}
+          showMax
+          required
+          warn={dustWarning}
+          help={(
+            <small>
+              Available on-chain balance: <Unit value={blockchainBalance} />
+              <a href="#"><Icon type="info-circle" /></a>
```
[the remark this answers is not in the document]
jamaljsr: Oh man, that slipped through the cracks. I’m working on the modal for that link in another branch and meant to take that out of this one. I’ll remove it

src/app/components/SendForm/ChainSend.tsx
jamaljsr: Bah, I think I know. The fee is required to calculate the dust threshold but the fee doesn’t get fetched until the Advanced button is clicked. This will be fixed once I make the fee always visable.

jamaljsr:
I made the updates based on the review. I think it should be good to go.

PR #169, opened 2019-02-16
Fix wrong name for value field in payments

### LINKED ISSUE #107: Non-existing transactions shown
### Description of the Issue

Within Transactions list, weird transactions are shown.

Expected behavior: show list of onchain and lightning transactions (both outgoing and paid invoices). On click, show transaction details in right pane.

Current behavior: 
Within transaction list, 3 separate outgoing lightning payments are shown to 2 nodes I did transactions with. All 3 have the same amount 3 244 570 453 sat. 
Upon click on tx, no tx detail pane is shown, and main pane with tx list dissapears. Side issue: wrong transaction date shown.

Check on my LND node:
`lncli listpayments` does not show such outgoing payments (fortunately).
`lncli listchannels` shows that channels open to concerned nodes have total capacity lower then transactions amount shown.

### Steps to Reproduce

Click on transaction list.

### Description of Your Setup

*Browser & version*: Chrome
*Operating system*: Windows 10
*Node software & version*: Remote node accessed through LAN running LND 0.5.0-beta

image

image

`lncli listpayments` response

image

alepop:
Closes #107

### Description

This pr solves problems with:
- wrong payments value for all payments
- problem with the extension crash after clicking to the payment transaction

### Steps to Test

1. Click `Transactions` tab
2. Click on `Payments` tx

### DIFF
--- a/src/app/modules/account/sagas.ts
+++ b/src/app/modules/account/sagas.ts
@@ -71,15 +71,26 @@ export function* handleGetTransactions(): SagaIterator {
     const paymentNodes: Array<Yielded<typeof nodeLib.getNodeInfo>> = yield all(
       paymentNodeIds.map(id => call(safeGetNodeInfo, nodeLib, id))
     );
-    const payments = paymentsRes.payments.map(p => {
-      const nodeResponse = paymentNodes.find(
-        n => p.path[p.path.length - 1] === n.node.pub_key
-      );
-      return {
-        ...p,
-        to: (nodeResponse as any).node,
-      };
-    });
+    const payments = paymentsRes.payments
+      .map(p => {
+        const nodeResponse = paymentNodes.find(
+          n => p.path[p.path.length - 1] === n.node.pub_key
+        );
+        return {
+          ...p,
+          to: (nodeResponse as any).node,
+        };
+      })
+      .map(p => {
+        if (!p.value_sat && !p.value_msat) {
+          return {
+            ...p,
+            value_sat: (p as any).value,
+            value_msat: `${(p as any).value}000`,
+          }
+        }
+        return p;
+      });
 
     yield put({
       type: types.GET_TRANSACTIONS_SUCCESS,


alepop:
Just found in the lnd API docs, that a `value` field is deprecated for `Payment`, but my `lnd` node with version `0.5.0-beta commit=3b2c807288b1b7f40d609533c1e96a510ac5fa6d` not return `value_sat` and `value_msat` fields in the `listpayments` responce. Anyway I will update this PR with backward compatibility for deprecated field.

wbobeirne:
Good catch, but to reduce the number of places that this might crop up, perhaps we should just reassign `.value` to `.value_sat` and `.value_msat` in the API handler? That way it'll be treated the same as though it sent those back.

alepop:
@wbobeirne pr updated

src/app/modules/account/sagas.ts
```diff
@@ -71,15 +73,26 @@ export function* handleGetTransactions(): SagaIterator {
...
+          ...p,
+          to: (nodeResponse as any).node,
+        };
+      })
+      .map(p => {
+        if (!p.value_sat && !p.value_msat) {
+          return {
+            ...p,
+            value_sat: p.value,
+            value_msat: fromBaseToUnit(p.value, Denomination.MILLIBITCOIN),
```
wbobeirne: msat is not millibitcoin, it's millisats, where each is one one-thousandth of a satoshi. We're not using it anywhere, but this would be an incorrect number.

As far as I can tell, the simplest but best way to convert value to this would be to do `${p.value}000`.

src/app/lib/lnd-http/types.ts
```diff
@@ -142,6 +142,7 @@ export interface LightningPayment {
   payment_hash: string;
   fee: string;
   creation_date: string;
+  value: string; // deprecated
```
wbobeirne: This should be removed now.

src/app/modules/account/sagas.ts
alepop: oops. fixed

alepop:
@wbobeirne pr updated with your suggestions

PR #170, opened 2019-02-17
Add Node Pubkey/URI to main page

### LINKED ISSUE #160: Show your node's pubkey (and connection info?)

Just now realizing there's no way to get your own node's pubkey in an easy copyable way. Perhaps clicking on your icon should show it? Bonus points for if we can expose the full connection string, e.g. `[pubkey]@ip:port`. Not all node setups will allow for connecting that way, but many should. 

#### Zap

screen shot 2019-02-05 at 9 12 49 pm

callmekurisu:
Closes #160 

### Display Node URI/pubkey

Upon clicking Identicon next to alias/pubkey display pubkey/URI for copying with QR code. Public node will display Node URI. Private node will display pubkey only.

### Steps to Test

1. Add ```externalip``` to lnd.conf if not already set
2. Restart lnd
3. Click Identicon next to alias/pubkey on main page
4. Repeat step 1-3 for private node with ```externalip``` not set in lnd.conf


src/app/components/AccountInfo/NodeUriModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+class NodeUriModal extends React.Component<Props> {
+
+  render() {
+    const {
+      node,
+      isOpen,
+      onClose
+    } = this.props;
+    const text = node!.uris[0] === undefined ? node!.identity_pubkey : node!.uris[0];
```
wbobeirne: `node` could be undefined here, that should probably be handled instead of ignored, even if it's just a `return null`. You also probably want to do a falsy check here instead of undefined, in case it got set to null.

src/app/components/AccountInfo/NodeUriModal.tsx
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { Modal } from 'antd';
+import { getNodeInfo } from 'modules/node/actions';
+import QRCode from 'qrcode.react';
+import Copy from 'components/Copy';
+import { AppState } from 'store/reducers';
+import './DepositModal.less';
```
wbobeirne: I don't think you want to import this

src/app/components/AccountInfo/NodeUriModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+interface StateProps {
+  node: AppState['node']['nodeInfo'];
+}
+
+interface DispatchProps {
+    getNodeInfo: typeof getNodeInfo;
```
wbobeirne: getNodeInfo is never used in this component

src/app/components/AccountInfo/index.tsx
```diff
@@ -46,7 +49,7 @@ class AccountInfo extends React.Component<Props, State> {
   render() {
     const { account, chain } = this.props;
     const { isDepositModalOpen } = this.state;
-
+    const { isNodeUriModalOpen } = this.state;
```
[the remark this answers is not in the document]
callmekurisu:
```Typescript 
render() {
    const { account, chain } = this.props;
    const { isDepositModalOpen,isNodeUriModalOpen } = this.state;

File: src/app/components/AccountInfo/NodeUriModal.tsx (hunk 1 above)
Comment (callmekurisu) [reply src/app/components/AccountInfo/NodeUriModal.tsx]:
```Typescript
const text = node === null ? '' : !!node.uris[0] === false ? 
  node.identity_pubkey : node.uris[0];
const title = node === null ? '' : !!node.uris[0] === false ? 
  'Pubkey' : 'Node URI';

File: src/app/components/AccountInfo/NodeUriModal.tsx (hunk 2 above)
Comment (callmekurisu) [reply src/app/components/AccountInfo/NodeUriModal.tsx]: Oops! Removed in d02a89f

File: src/app/components/AccountInfo/NodeUriModal.tsx (hunk 3 above)
Comment (callmekurisu) [reply src/app/components/AccountInfo/NodeUriModal.tsx]: Dang, I'm slippin. Forgot to put {} when there is no DispatchToProps. Fixed in d02a89f.

**Comment by callmekurisu** [comment]
I got prettier on VS code. I don't think I'm using it correctly. Should be ctrl+shift+i. Will work on that. 
Did some cleanup. Let me know if I missed anything! Thanks for the review!

File: src/app/components/AccountInfo/NodeUriModal.tsx (hunk 3 above)
Comment (callmekurisu) [reply src/app/components/AccountInfo/NodeUriModal.tsx]: Dang, I'm slippin. Forgot to put {} when there is no DispatchToProps. Fixed in

File: src/app/components/AccountInfo/NodeUriModal.tsx [hunk 5]
```diff
@@ -0,0 +1,54 @@
...
+  onClose(): void;
+}
+
+type Props = StateProps & OwnProps;
+
+class NodeUriModal extends React.Component<Props> {
+  render() {
+    const { node, isOpen, onClose } = this.props;
+    const text =
+      node === null ? '' : !!node.uris[0] === false ? node.identity_pubkey : node.uris[0];
```
wbobeirne: When I mentioned a falsy check, I didn't mean literally `=== false`. You want to catch all sorts of cases, false, undefined, null, empty string etc. So in those cases, you just do `node && node.uris && node.uris[0] ? /* true case here */ : /* false case here */`. That catches any instance where node or node.uris[0] isn't what we want.

callmekurisu: Ahh, I see. My bad. (^_^)
```Typescript
const text = node === null ? '' :
      node && node.uris && node.uris[0] ?
      node.uris[0] : node.identity_pubkey;
const title = node === null ? '' :
      node && node.uris && node.uris[0] ?
      'Node URI': 'Pubkey';

PR #172, opened 2019-02-18
Refactor Onboarding navigation

PJUllrich:
# DO NOT MERGE (yet)

This is only a first draft of a refactor attempt to use React Router instead of state checks in order to navigate through the onboarding process.

- I changed all components in `SelectNode` and `Splash`.
- I added new routes to `AppRoutes`. The routes need to be here since React Router does not work with Nested Routes in Chrome Extensions.

### Steps to Test

Just delete all settings and go through the onboarding process.

### Still to do / to discuss
- Implement the `Lightning App` button.
- Test the `BTCPayServer` button.
- By using `this.props.history.push` in the `SelectNode` components, the flow between components is hard-coded. Sometimes, these components are used individually (e.g. in the Settings page). We have to find a way to adjust the path to which a component navigates.
- Regarding the transition from `UploadMacaroon` to `ConfirmNode`: Should we call `setNode` upon exiting `UploadMacaroon`, or within `ConfirmNode`? In case of the later, we need to somehow pass on the info from `UploadMacaroon` to `ConfirmNode`, which is not optimal.

### Screenshots

short


src/app/components/SelectNode/BTCPayServer.tsx
```diff
@@ -12,80 +16,101 @@ export interface BTCPayServerConfig {
...
-        </ol>
-        <Form className="BTCPayServer-form" onSubmit={this.handleSubmit} layout="vertical">
-          <Form.Item
-            label="Connect data"
-            help={error}
-            validateStatus={error ? 'error' : url ? 'success' : undefined}
-          >
-            <Input
+      <div>
+        <h2 className="SelectNode-title">Connect to your BTCPay Server</h2>
```
wbobeirne: It might just be better to make a `<SelectNodeTemplate title="Title Goes Here">` component than having to implicitly copy the styles and classnames across 5 components. That way future changes (e.g. back button) would only require changes in one place.

src/app/AppRoutes.tsx
```diff
@@ -27,49 +33,105 @@ const routeConfigs: RouteConfig[] = [
...
+    route: {
+      path: '/onboarding-node-confirm',
+      component: ConfirmNode
+    },
+    template: {}
+  },
+  {
+    // Onboarding
+    route: {
+      path: '/onboarding-password',
```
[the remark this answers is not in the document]
PJUllrich: I agree. I tried e.g. `/onboarding/password` first, but `React Router` didn't work with these subpaths. That's why chose the `-` combination.

src/app/AppRoutes.tsx
```diff
@@ -27,49 +33,105 @@ const routeConfigs: RouteConfig[] = [
     // Initial loading
     route: {
       path: '/',
-      render: () => <Loader />,
-      exact: true,
+      render: () => <Loader/>,
+      exact: true
     },
-    template: {},
+    template: {}
```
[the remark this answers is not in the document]
PJUllrich: Yes, I'll stop the auto-formatting.

src/app/AppRoutes.tsx
```diff
@@ -27,49 +33,105 @@ const routeConfigs: RouteConfig[] = [
...
   {
     // Onboarding
     route: {
       path: '/onboarding',
-      component: OnboardingPage,
+      component: OnboardingPage
     },
     template: {
-      hideHeader: true,
+      hideHeader: true
```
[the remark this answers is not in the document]
PJUllrich: I wasn't aware what this `hideHeader` attribute was for. I would add it to the sub-pages of onboarding if we decide to go further down this route of using React Router.

src/app/components/SelectNode/ConfirmNode.tsx
```diff
@@ -1,57 +1,94 @@
...
+}
+
+type Props = StateProps & DispatchProps & RouteComponentProps;
+
+class ConfirmNode extends React.Component<Props> {
   render() {
-    const { nodeInfo, onConfirm, onCancel } = this.props;
-    const chain = nodeInfo.chains[0] as CHAIN_TYPE;
+    const {nodeInfo} = this.props;
+    const chain = nodeInfo!.chains[0] as CHAIN_TYPE;
```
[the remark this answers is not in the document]
PJUllrich: I actually wanted to do that, but I didn't know how to do that and couldn't find anything online. Thank you for the answer I've been looking for!

src/app/components/SelectNode/BTCPayServer.tsx
PJUllrich: True that. Good call.

PJUllrich:
I agree with you that this solution using React-Router doesn't cut it. Especially since it is awkward using React Router in a Chrome Extension, where some weird things just don't work (like having `Route` components in sub-components. I suggest that I will look into HashRouters and MemoryRouters and see whether they offer a better solution.

src/app/AppRoutes.tsx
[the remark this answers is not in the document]
PJUllrich: I updated this in the latest commit.

src/app/components/SelectNode/ConfirmNode.tsx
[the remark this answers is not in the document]
PJUllrich: This was handled with an `if ... else ... ` statement in the latest commit.

src/app/components/SelectNode/BTCPayServer.tsx
PJUllrich: This is handled in the latest commit by using a `<TitleTemplate>` instead of `<h2>` elements.

PJUllrich:
I updated this PR. Here is what has changed:
- I removed all sub-paths from `AppRoutes` and moved them as `<Route>`-elements to `OnboardingPage`.
- I removed all hard-coded forward navigation from all sub-components of `SelectNode`.
- I added a `onComplete` function to all sub-comps of `SelectNode` that is called ... surprise ... whenever the sub-component has completed what they are supposed to do.
- I moved the `<h2>` title from all sub-components to a `TitleTemplate` component.
- I gave the `TitleTemplate` component a back button for navigation purposes. 
- I added a `LightningApp` component, that checks connection with `lightning-app` and navigates forward upon success.

wbobeirne:
Sorry for leaving this hanging so long, I had some travel that kept me away from the computer!

Unfortunately, in the latest version, it looks like resetting node settings from the popup has broken. I believe this has to do with the router redirecting you to `/onboarding`, which closes the popup before the settings sync, presenting you with the page still logged in.

I'll manually reset my local storage settings and continue to test, but I thought I'd drop that note here first!

src/app/components/Splash/index.tsx
```diff
@@ -4,23 +4,18 @@ import { browser } from 'webextension-polyfill-ts';
...
-    if (process.env.APP_CONTAINER === 'page') {
-      browser.storage.local.get('skipSplash').then(value => {
-        if (value && value.skipSplash) {
-          browser.storage.local.remove('skipSplash').then(() => {
-            this.handleContinue();
-          });
-        }
-      });
+    if (process.env.APP_CONTAINER !== 'page') {
+      browser.runtime.openOptionsPage().then(window.close);
```
wbobeirne: While I don't like adding unnecessary steps for users, I don't like that clicking the icon initially causes the options page to open. I want users to know that clicking the icon will open a popup and not interrupt their current session (normally) and that onboarding is a special case. Reverting this should also fix the issue mentioned in my previous comment above.

src/app/components/SelectNode/TitleTemplate.tsx
```diff
@@ -0,0 +1,34 @@
...
+  };
+
+  render() {
+    const {title} = this.props;
+    return (
+      <div className="TitleTemplate">
+        <div className="TitleTemplate-icon-back">
+          {this.props.showBack &&
+          <Icon type="left" onClick={this.props.history.goBack}/>}
+        </div>
```
wbobeirne: This should be made a button or a tag for accessibility and cursor behavior.

src/app/components/SelectNode/LightningApp.tsx
```diff
@@ -0,0 +1,114 @@
...
+                   target="_blank">
+                  clicking this link
+                </a>{' '}
+                and making sure it loads correctly. If there are SSL errors,
+                click "advanced" and proceed to accept the certificate.
+              </p>
+            </>}
+            showIcon
+            closable
+        />
```
wbobeirne: Might be good to componentize this since it's copy / pasted from InputAddress, and will likely be reused for a Zap onboarding component.

src/app/components/SelectNode/UploadMacaroons.tsx
```diff
@@ -1,171 +1,216 @@
...
-                Upload <code>readonly.macaroon</code>
-              </p>
-              <p className="ant-upload-hint">
-                Click or drag to upload
-              </p>
-            </Upload.Dragger>
-
-            {dirs &&
+      <div>
+        <TitleTemplate title={'Upload Macaroons'}/>
```
wbobeirne: Unfortunately the back button here doesn't work with LightningApp since it auto-connects. Might be good to have the back button here take you back to the selection of node type, rather than whatever the previous page was.

src/app/components/SelectNode/UploadMacaroons.tsx
```diff
@@ -1,171 +1,216 @@
...
-        {error &&
+          {error || checkAuthError &&
           <Alert
-            message="Invalid macaroon"
-            description={error}
-            type="error"
-            closable
-            showIcon
+              message="Invalid macaroon"
+              description={error || checkAuthError}
```
wbobeirne: It looks like this is causing it to try rendering an `Error` object which causes the following error in React:
```
Uncaught Invariant Violation: Objects are not valid as a React child (found: Error: Read-only macaroon did not authenticate). If you meant to render a collection of children, use an array instead.
```
The screen blanks out when this happens. I think you want the `.message` attribute instead.

PR #177, opened 2019-02-24
Add new screen to display wallet balance sources

### LINKED ISSUE #31: Send bitcoin transactions

jamaljsr:
Closes #31 (already closed)
I originally intended to include this feature with the on-chain sending but I knew this code change was going to be too big, so here's a separate PR.

### Description

This is my attempt at trying to answer the question "Why are my balances not adding up?" which many users have asked. The typical reason is simply because you opened a channel of a small size but LND used a UTXO which is much larger, so until the channel is opened you cannot spend the full amount of the UTXO. 

This screen also surfaces some other useful information such as:
- I have a channel open with an offline node, when were they last online?
- How long will it take for my pending balance to become available?
- What UTXOs do I have in my wallet?
- When will this force closed channel be confirmed?

### Steps to Test

1. View the Balances screen and confirm it looks accurate for your node
2. Open & Close channels to see the pending balances
3. Send & Receive on-chain coins to see your UTXOs get updated


src/app/components/Balances/index.tsx
```diff
@@ -0,0 +1,217 @@
+import React from 'react';
+import { connect } from 'react-redux';
+import { Tooltip, Progress, Tabs, Icon, Select, Alert } from 'antd';
+import Statistic from 'antd/lib/statistic';
```
wbobeirne: Very cool, did not know about this component!

src/app/components/Balances/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+            <Statistic 
+              title="Lightning" 
+              value={parseFloat(fromBaseToUnit(stats.channelTotal, denomination))} 
+              valueStyle={{ color: '#7642ff' }} 
+            />
+            <Statistic 
+              title={blockchainDisplayName[chain]}
+              value={parseFloat(fromBaseToUnit(stats.onchainTotal, denomination))} 
+              valueStyle={{ color: '#ff9500' }} 
+            />
```
wbobeirne: Would these guys with suffixes not fit? Looks a little awkward having numbers without indicators of what they are.

src/app/components/Balances/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+              >
+                <BalanceDetails
+                  groups={stats.channelDetails} 
+                  chain={chain}
+                  denomination={denomination}
+                  emptyMessage="Open some channels to enjoy the Lightning experience"
+                />
+              </Tabs.TabPane>
+              <Tabs.TabPane
+                tab={<><Icon type="link"/> Bitcoin</>}
```
wbobeirne: Should use `blockchainDisplayName` for multiple chain compatibility

src/app/components/Balances/index.tsx
```diff
@@ -0,0 +1,217 @@
...
+                key="on-chain"
+              >
+                <BalanceDetails 
+                  groups={stats.onchainDetails} 
+                  chain={chain} 
+                  emptyMessage="Fund you"
```
wbobeirne: Maybe I'm missing context but is this message supposed to be more descriptive?

src/app/lib/lnd-http/index.ts
```diff
@@ -360,6 +360,16 @@ export class LndHttpClient {
     );
   };
 
+  getUtxos = (params: T.GetUtxosParams = { max_confs: 100000000 }) => {
+    // max_confs must be set to a number higher than the chain's current
+    // block height in order to get all utxos for a wallet
```
wbobeirne: Interesting quirk, what if it's not sent? What's the default behavior?

src/app/components/Balances/index.tsx
jamaljsr: Yeah, I discovered it by looking at the antd pro demo site

src/app/components/Balances/index.tsx
jamaljsr: I added the suffixes but needed to do a bit of trickery with the styles in order to make the content fit depending on how big the numbers are.

src/app/components/Balances/index.tsx
jamaljsr: Darnit! I thought I got all of those hard-coded references. Good catch

src/app/components/Balances/index.tsx
jamaljsr: Yup.. it was probably late and I lost my train of thought.

src/app/lib/lnd-http/index.ts
jamaljsr: Very strange indeed, I thought the same as well. If you don't supply a value it will default to 0 and only return UTXOs that have 0 confs. I picked this huge number cause the block_height of BTC/LTC won't get this high for many years to come

jamaljsr:
@wbobeirne Thanks for the review and catching those oversights of mine. I gotta remind myself to go back and review with fresh eyes a day or two later. Not sure how I missed those obvious issues.

wbobeirne:
'Sall good dude! Gonna run through one more time then merge this bad boy.

wbobeirne:
Excellent changes, looks & works great. Merging!

PR #179, opened 2019-03-04
Display channel details when clicked on the home page

jamaljsr:
Based on #177 so it would be best to merge that PR first as those commits are included in this branch.

### Description

Adds a new slide out drawer with more details about a channel.

_Note: This PR also includes code to implement closing channels but the button has been hidden. There is an open issue with LND [#2730](https://github.com/lightningnetwork/lnd/issues/2730) which prevents closing channels via the REST API. When that issue gets resolved, we can show the button to enable closing channels. If it's preferred to remove this code, I can do that. I just felt it doesn't hurt to leave it there even if it's currently not used._

### Steps to Test

1. Open Joule
2. Click on one of the channels listed on the home screen
3. Review the details in the slide-out drawer that opens


wbobeirne:
I'll wait for #177 to be merged before reviewing, to avoid double-reviewing the same code.

wbobeirne:
@jamaljsr #177 is in, would you mind merging `develop` into this branch to fix the PR diff?

jamaljsr:
Yup, will do it in an hour or so. Thanks

jamaljsr:
I merged in the develop branch

src/app/components/ChannelInfo/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+            <div className="ChannelInfo-parties-party-name">
+              {channel.node.alias || channel.node.pub_key}
+            </div>
+          </div>
+        </div>
+
+        <table className="ChannelInfo-details">
+          <tbody>
+            {this.getChannelDetails().map(d => (
+              <tr className="TxInfo-details-row" key={d.label}>
```
wbobeirne: Looks like these class names are off. This could probably be spun out into its own component, since as seen here, it's very reusable! Your call on whether you want to componentize or just change the class names.

src/app/components/ChannelInfo/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+        label: 'Closing Tx',
+        value: txLink(channel.closing_txid),  
+      });
+    } else if (channel.status === CHANNEL_STATUS.FORCE_CLOSING) {
+      details.push({
+        label: 'Closing in',
+        value: <>
+          <span>{channel.blocks_til_maturity} blocks</span>
+          <span className="Unit-secondary">
+            ~{moment().add(channel.blocks_til_maturity * 10, 'minutes').fromNow(true)}
```
wbobeirne: Awesome, love how much easier this is to grok than blocks.

src/app/components/ChannelInfo/index.tsx
```diff
@@ -0,0 +1,196 @@
...
+            {this.getChannelDetails().map(d => (
+              <tr className="TxInfo-details-row" key={d.label}>
+                <td className="TxInfo-details-row-label">{d.label}</td>
+                <td className="TxInfo-details-row-value">{d.value}</td>
+              </tr>
+            ))}
+          </tbody>
+        </table>
+
+        {/* TODO: show button when LND issue #2730 is resolved */}
```
wbobeirne: Thanks for leaving this in, should be simple to hook up once this is fixed or we figure out way to switch to gRPC.

jamaljsr:
Thanks for the review. You can merge and fix if you have time tonight, otherwise I will be available to do it tomorrow. Either way is cool with me.

jamaljsr:
I refactored the icons and table into separate components which reduced a bunch of duplicate CSS. I was able to completely remove the stylesheet for the new `ChannelInfo` component. Thanks for the nudge, this is much better.

wbobeirne:
Awesome, thanks for going the extra mile!

PR #182, opened 2019-03-09
Add fee selection to Open Channel form

### LINKED ISSUE #181: Add fee selection to Open Channel form

This issue was brought up in this tweet: https://twitter.com/citlayik/status/1103914960952815617
It suggests that LND can overpay transaction fees at times. I have been unable to confirm this myself, but it doesn't hurt to give users more control. Since we recently introduced fee selection for sending on-chain transactions, it makes sense to also add it for opening a channel as well.

jamaljsr:
Closes #181 

### Description

Adds a new field to the Open Channel form giving the option to choose the transaction speed. This will set the sats/byte to a predetermined value based on the current BTC mainnet mempool.

_Note: Fee estimates are only provided for BTC so this field is hidden when connected to an LTC node_

### Steps to Test

1. Click on `Open channel` from the top right nav menu
2. Click on the `Show advanced fields` button in the modal
3. Choose a fee level (auto/slow/normal/fast)
4. Confirm the sats/byte changes in the help message below the toggle
5. Click the `Open Channel` button to send the transaction
6. View the transaction in a block explorer to verify that your fee was used. You can open two channels to to confirm that choosing the Fast option is cheaper than Auto


jamaljsr:
I added the ability to manually specify a custom fee (sats/byte) in the fee selection field. It also displays a warning if the specified fee is larger than the fastest recommended fee.

jamaljsr:
I just committed an alternative UI using a slider instead of the toggle/input combo. I think this may be a better solution as it reduces the chances of someone fat-fingering a really high fee unnecessarily. It also greatly simplifies the error detection code.



If you prefer the previous UI, then I can easily revert the last commit. I just wanted to throw this out there as I felt it is a better approach.

wbobeirne:
Love the new UI. Will be taking a look at this a little later, but definitely prefer this direction!

src/app/components/OpenChannelModal/index.tsx
```diff
@@ -266,6 +277,8 @@ class OpenChannelModal extends React.Component<Props, State> {
         address,
         capacity,
         isPrivate,
+        // only send non-zero fee
+        fee: fee ? fee.toString() : undefined,
```
wbobeirne: Does `handleFeeSubmit` in `ChainSend` need the same consideration here? Looks like it could send 0 right now.

src/app/components/FeeSelectField/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+              0: 'auto',
+              [onChainFees.hourFee]: '',
+              [onChainFees.halfHourFee]: '',
+              [onChainFees.fastestFee]: 'fastest',
+            }}
+            tipFormatter={(v) => {
+              switch (v) {
+                case onChainFees.fastestFee: return 'fastest';
+                case onChainFees.halfHourFee: return 'normal';
+                case onChainFees.hourFee: return 'slow';
```
wbobeirne: Would be nice if these kept the number as well, e.g. `16 (normal)`

jamaljsr:
Great feedback. Thanks for the review. I’ll get these couple items fixed up as soon as I can.

src/app/components/FeeSelectField/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+class FeeSelectField extends Component<Props, State> {
+  state: State = {
+    fee: 0,
+  }
+
+  componentDidMount() {
+    this.props.getOnChainFeeEstimates();
+  }
+
+  componentWillReceiveProps(nextProps: Props) {
```
otech47: FYI `componentWillReceiveProps` will be deprecated in React 17 and generally a good upgrade path for this kind of function is to use `componentDidUpdate` instead and compare against `prevProps` instead of `nextProps`

https://reactjs.org/docs/react-component.html#updating

More of a long term consideration but noticed it because I write a ton of React and have gotten used to it by now

src/app/components/FeeSelectField/index.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  componentDidMount() {
+    this.props.getOnChainFeeEstimates();
+  }
+
+  componentWillReceiveProps(nextProps: Props) {	
+    const { onChainFees } = this.props;	
+    if (nextProps.onChainFees !== onChainFees && nextProps.onChainFees !== null) {	
+      const { fastestFee } = nextProps.onChainFees;
+      this.setState({ fee: fastestFee });
```
otech47: Also just took a look at the parent component and am wondering whether you wouldn’t need local state in this component and you can just pass the `fee` as a prop from the `ChainSend` component since it seems like you’re managing changes to both this local state and the parents local state?

All good if you have your reasoning that I don’t see, but in any case these are just architecture suggestions and not functionally critical. 

Love the slider UI btw 👍👍👍

jamaljsr:
hey @wboeirne I am traveling and wont be able to make any changes to this branch until this weekend. Feel free to make any additional updates you think are improvements. 

Great feedback @otech47, thank you very much.

wbobeirne:
No worries, hope it's somewhere sunny. I just pushed changes to address the feedback, and merged develop back in to fix the snyk CI. If you don't get a chance to review, I'll probably merge as is since I tested it pretty thoroughly and it all LGTM.

jamaljsr:
@wbobeirne Thanks for cleaning this up and getting it merged 👍

PR #205, opened 2019-04-24
Fixed Connection problem URL formats  #198

### LINKED ISSUE #198: Connection problem URL formats with double-slash
### Description of the Issue

Got a message about this:

> just an FYI, when I click on the connection problem link fixer thing in the 0.5.0 version it puts in a double slash... `https://localhost:3000//v1/getinfo`

Requires a little further investigation.

### Steps to Reproduce

?

callmekurisu:
Closes #198 

### URL double-slash when adding slash during initial setup

Modified InputAddress.tsx 'submittedUrl' value in the state. The submitted URL is decomposed and only the host and port are extracted. Any additional characters after the port number are removed and joined with remaining test URL /v1/getinfo.

### Steps to Test

1. Clear browser history
2. Begin to setup the Joule Extension
3. Enter URL with trailing foward slash on the Failed to connect screen
4. Verify no additional slashes are added

### DIFF
--- a/src/app/components/SelectNode/InputAddress.tsx
+++ b/src/app/components/SelectNode/InputAddress.tsx
@@ -28,12 +28,6 @@ export default class InputAddress extends React.Component<Props, State> {
     const { error, isCheckingNode } = this.props;
     const { validation, url, submittedUrl } = this.state;
     const validateStatus = url ? validation ? 'error' : 'success' : undefined;
-    // Handle accidental user input after port number
-    const modUrl = submittedUrl.split(":");
-    // Basically just break apart the URL
-    const modPort = modUrl[2].split("/");
-    // Eliminate anything after port number and bring back together
-    const altUrl = `${modUrl[0]}:${modUrl[1]}:${modPort[0]}`;
     const help = (url && validation) || (
       <>
         You must provide the REST API address. Must begin with{' '}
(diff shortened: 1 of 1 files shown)


wbobeirne:
Thanks for taking a crack at this one! I'm not sure I'm ok with this fix though, as it makes some assumptions about user's node setups that I'm not willing to make. If someone's node is available at a path, that should be acceptable. I'd rather just normalize URLs by chopping off any trailing slashes.

wbobeirne:
See where this is handled in the http api lib: https://github.com/joule-labs/joule-extension/blob/develop/src/app/lib/lnd-http/index.ts#L14-L16

callmekurisu:
@wbobeirne ah I see. The regex solution is makes much more sense. So just remove the trailing slash from user input and submit that, yeah?

```javascript
private handleSubmit = (ev: React.FormEvent<HTMLFormElement>) => {
    const url = this.state.url.replace(/\/$/, '')
    ev.preventDefault();
    browser.permissions.request({
      origins: [urlWithoutPort(this.state.url)],
    }).then(accepted => {
      if (!accepted) {
        message.warn('Permission denied, connection may fail');
      }
      this.props.submitUrl(url);
      this.setState({ submittedUrl: url });
    });
```

PR #207, opened 2019-04-24
Fix pathless transactions crashing tx list

### LINKED ISSUE #206: Transactions tab crashes for some nodes
### Description of the Issue

Getting a report of the transactions page not loading, with an error of "Cannot read property length of undefined". Narrowed it down to the following but of code:

https://github.com/joule-labs/joule-extension/blob/develop/src/app/modules/account/sagas.ts#L68-L70

Looks like sometimes payments come back with an undefined (or empty, due to the rest rpc bug) path array, so we'll just need to add it to the RPC call default properties:

https://github.com/joule-labs/joule-extension/blob/develop/src/app/lib/lnd-http/index.ts#L183

### Steps to Reproduce

Unknown, but can be fixed without reproducing

wbobeirne:
Closes #206

### Description

Fixes some payments (I think only quite old ones) from crashing the transactions tab. This happened due to them coming back with `undefined` paths. In total, this PR does the following:

* Defaults the `path` array to `[]` if undefined
* Improves the display of pathless payments
* Improves the Identicon component by having it display a (?) when the pubkey is falsy

### Steps to Test

1. Get a pathless payment, or alter the code so that you can forcefully induce one
2. Load the transactions tab, confirm it doesn't crash
3. Open a pathless transaction, confirm it displays correctly

### List view

Screen Shot 2019-04-24 at 2 26 18 PM

### Detail view

Screen Shot 2019-04-24 at 2 28 40 PM

### DIFF
--- a/src/app/components/Identicon.tsx
+++ b/src/app/components/Identicon.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import classnames from 'classnames';
 import { toSvg } from 'jdenticon';
+import { Icon } from 'antd';
 import './Identicon.less';
 
 interface Props extends React.HTMLProps<HTMLDivElement> {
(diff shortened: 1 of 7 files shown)


wbobeirne:
Confirmed working here: https://github.com/joule-labs/joule-extension/issues/206#issuecomment-486878728

Merging in!

PR #209, opened 2019-04-24
Sync warning

### LINKED ISSUE #199: Show banner when synced_to_chain is false

Noticed this while watching someone upgrade today, if your getInfo response has `synced_to_chain: false`, we should display a banner warning you that you are syncing and may be missing funds temporarily. This can avoid a lot of user confusion.

wbobeirne:
Closes #199

### Description

Adds a banner when your node is still syncing to the chain to let you know that your balances might be off. To my knowledge there's no way to get progress, since it always reports the bitcoin node's block height, not LND's current scan progress, so the best we can do is this boolean display.

### Steps to Test

Open joule with a syncing node (or induce the state yourself programmatically), confirm you see a banner.

## Screenshots

Screen Shot 2019-04-24 at 5 02 08 PM

### DIFF
--- a/src/app/components/AccountInfo/index.tsx
+++ b/src/app/components/AccountInfo/index.tsx
@@ -9,18 +9,21 @@ import Unit from 'components/Unit';
 import DepositModal from './DepositModal';
 import NodeUriModal from './NodeUriModal';
 import { getAccountInfo } from 'modules/account/actions';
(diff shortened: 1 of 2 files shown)


callmekurisu:
Verified. Nice work!
 
image

wbobeirne:
Excellent, thanks for giving it a look.

PR #210, opened 2019-04-26
Run & enforce prettier with CI & precomit hook

wbobeirne:
No issue related

### Description

Runs prettier across the codebase, and enforces it with a precommit hook via husky as well as checks against it in CI.

I'm not 100% wild about all of its changes, some lines have gotten more verbose (Object arguments, "flat" style arrays) while others are a little more obfuscated (Ternaries, components that came right up against char limit that got flattened) but I think consistency is king here. It'll make submitting PRs much less of a hassle if there are significant code style changes, or the user's own formatter ran amok.

I'll leave this open for a bit in case anyone has any strong opinions about some of the changes.

### Steps to Test

1. Write ill-formatted code
2. Commit said code
3. Confirm it's now correctly formatted


jamaljsr:
#winning

Sometimes prettier does things I don’t like but i certainly prefer consistency over personal preferences when it comes to code style in a team environment. For an open source repo, its a must-have. Great addition to the project

wbobeirne:
Glad you're into it! Thanks for the catch, too. I think I had an old version of husky in my node_modules that was keyed off of the style. Updated with yours.

PR #220, opened 2019-06-04
Provide private channel route hints & make private routes sendable

### LINKED ISSUE #192: Handle private channels & route hints

Before sending payments, we run a query of the routes to make sure it's possible and to get a fee estimate. However, we don't use the routing hints from the invoice, so routing to nodes with only private channels always fails. This means it's impossible to send to private channel only nodes, and any request that includes hints will be ignored. These should be included in the route query.

Likewise, when generating invoices, we don't pass the `private` boolean to include routing hints. We should do this by default, with a checkbox to disable it.

wbobeirne:
Closes #192 

### Description

This makes invoice creations include private route hints, and also works around the issue where you were unable to send invoices that relied on private routes due to QueryRoute's inability to take in routing hints, and Joule's running of QueryRoutes before letting a user send to ensure that there _is_ a route and to provide the user fee information before they send.

The latter part of this should be moot as soon as this PR is merged https://github.com/lightningnetwork/lnd/pull/2186 but even then will stay an issue for folks running older node versions for a while. This janky temporary fix of returning placeholder routing info will probably have to sit in for a while until there's a reasonable expectation that most people have upgraded.

I also make a `Help` component and used it in a couple spots to explain some stuff.

### Steps to Test

1. Have a private channel open on your node
2. Generate an invoice with the routing hints checkbox checked
3. Run your invoice through https://lndecode.com/ and confirm the `routing_information` info is provided and correct
4. Generate an invoice with the routing hints checkbox unchecked
5. Run it through lndecode again, confirm the `routing_information` info is not provided
6. Attempt to pay the invoice you generated in step 2, confirm that you are able to attempt a send in spite of the fact that QueryRoutes failed
7. Attempt to pay the invoice you generated in step 4, confirm that you are unable to attempt a send since it has no routing information

### Invoice form w/ checkbox

Screen Shot 2019-06-04 at 1 45 48 PM

### Help Tooltip

Screen Shot 2019-06-04 at 1 45 55 PM

### Invoice prompt with advanced expanded

Screen Shot 2019-06-04 at 1 46 13 PM

### Display of fees with private routing

Screen Shot 2019-06-04 at 1 47 05 PM

### DIFF
--- a/src/app/components/ChannelList/ChannelRow.less
+++ b/src/app/components/ChannelList/ChannelRow.less
@@ -30,14 +30,18 @@
       }
     }
 
-    &-status {
+    &-status,
+    &-private {
       position: absolute;
       bottom: 0.1rem;
-      right: 0.1rem;
-      width: 0.8rem;
-      height: 0.8rem;
+      width: 0.9rem;
+      height: 0.9rem;
       border-radius: 100%;
-      box-shadow: 0 0 0 3px #FFF;
+      box-shadow: 0 0 0 3px #fff;
+    }
+
+    &-status {
+      right: 0.1rem;
 
       &.is-open {
         &.is-active {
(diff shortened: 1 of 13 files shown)


jamaljsr:
Nice! I'll try it out later today 👍

wbobeirne:
Have had this for a few weeks in testing and haven't had any issues pop up, so I'm just gonna merge this boyo in.

Issue #106, opened 2018-12-21
Consider native app companion


wbobeirne:
### Description of the Feature or Idea

This is kind of a "big picture" issue that would need to be broken out into its own repo or set of issues, but I wanted to open it for discussion, so here goes:

Joule is running into a lot of usability issues that are constraints from its extension form factor. If there were a native application, it could get around a bunch of issues:

* No more invalid SSL certificates since I can accept and use TLS cert files
* No more CORS issues and the extension permissions required to circumvent them
* Provide a node for the user if they don't already have one (neutrino, or offer full sync ala Pierre's node launcher)
* Scan common paths for automatic node configuration and macaroon lookup
* Alternative nodes that don't have REST APIs

This opens up a handful of questions though:
* Will the stand-alone extension still be supported? This might be a huge pain, but ideally people who want Joule with their existing node and who know what they're doing shouldn't need to download a whole app for it.
* Is there any security concern with having a native application?

Open to any thoughts on the matter!

### Existing Example(s) of Feature

* https://github.com/lightninglabs/lightning-app
* https://zap.jackmallers.com/

jnewbery:
My immediate reaction is "do we need another desktop lightning wallet?" There's already the [lightning labs app](https://github.com/lightninglabs/lightning-app) and [zap](https://zap.jackmallers.com/) for lnd, [spark](https://github.com/shesek/spark-wallet) for c-lightning, and [eclair](https://github.com/ACINQ/eclair)'s GUI. It's already quite a crowded field.

The beauty of lightning joule as it exists now is that the user flow is so smooth and intuitive, especially when the site integrates with webln. That's something that Joule offers that none of the other wallets have. Could you recreate that if Joule was a separate app?

bitcoinuser:
I agree that we don't need another desktop lightning wallet.

My goal use case with Joule is:

*  Pair the extension with an Android Wallet that support bluetooth.
* Use the great webln feature to pay with lightning.

jamaljsr:
I also have to agree that the biggest selling point of Joule for me is browser integration. The space needed a MetaMask for lightning/bitcoin and this extension is filling that hole. Also considering the potential future capabilities of streaming payments for websites, I think you have a winner here.

wbobeirne:
Thanks for the feedback y'all, this is exactly what I was hoping to get. Let me address some of your notes:

> Could you recreate that if Joule was a separate app?
> 
> [...]the biggest selling point of Joule for me is browser integration

[...]

PR #222, opened 2019-06-13
[WIP] Implement allowances that auto-pay invoices

### LINKED ISSUE #42: Allow user to setup automatic payments per lapp
Provide an interface where a user can setup automatic no-ask payments for certain lapps. This would be limited by domain, and have the following configurations:

* Maximum amount per transaction (e.g. ask if it's more than 1000 satoshis)
* Maximum amount total (e.g. up to 100,000 satoshis)
* Maximum amount per hour / day (e.g. 1000 satoshis per hour)

Ideally they could configure this during or after a payment request (Maybe a checkbox?) There should also be a list of active auto pay configs, and some UI treatment for when you're on a page that can auto-pay (and perhaps after each auto-payment.)

wbobeirne:
Closes #42

### Description

This adds a new module, `appconf` that can be used to add customizations to each webapp. The one implemented here is allowances, the idea of setting up rules that allow automatic payments to happen without user intervention.

This also adds a status bar that gives you quick actions you can take on the current page you're on.

A more thorough write-up of this PR is up on Medium: https://medium.com/p/2b08bec75e3a

### TODOs

* Make more chain agnostic (hard-coded use of `sats` in a few places)
* Find good places to provide fiat equivalents
* ???

### Steps to Test

1. Go to a WebLN page that you've enabled
2. Confirm you see the enabled status bar
3. Click the disable button, confirm it disables WebLN for app
4. Re-enable it, and click on the piggy bank, confirm it created a default allowance for you
5. Attempt to make a payment that fits the constraints, confirm it happened automatically for you
6. Attempt to make a payment that does _not_ fit the constraints, confirm you were prompted instead
7. Configure the settings and make sure they all work as expected
8. Add a new allowance directly from the allowances UI, confirm it correctly works when manually entered

### Allowance page

Screen Shot 2019-06-06 at 5 42 29 PM

### Notifications

Screen Shot 2019-06-06 at 5 40 31 PM

Screen Shot 2019-06-06 at 5 40 38 PM


bumi:
I was experimenting a bit with this PR. First of all: the UX is great! Super intuitive and nice to use. Can not wait to get this released :) 

Here are a few things that I've noticed: 

## send payment for allowances is done in the context of the website 
It seems currently the send payment call (POST /v1/channels/transactions) is done in a content script thus in the context of the current website. This does not work for me because of blocked cross origin calls. (calling the rest API of a lnd node without CORS). ([debug screenshot](https://twitter.com/Bumi/status/1143092468633808896/photo/1))
Somehow I feel like the check if it is a "allowed" domain and handling the request could be done in the same "handlePrompt" receiver? 

## failed browser notifications 
For whatever reason `browser.notifications` is undefined and thus `browser.notifications.create` and `browser.notifications.clear` fails. ([screenshot](https://twitter.com/Bumi/status/1143040537278189568/photo/1) 
I could not find out why this is the case. (I was using Chrome 75.0.3 on Linux)

## bolt11 (and coinfo/bitcoinjs-lib) currently does not support simnet network
And the bolt11 package adds some dependencies and I was wondering if we can use the `/v1/payreq` API call to decode the payment request.

wbobeirne:
> send payment for allowances is done in the context of the website

Yep, this is definitely a hacked together version. I think I'm going to have to setup a communication channel with the `background` script and do all node interactions in there. I definitely don't want node communications outside of the extension sandbox. Thanks for noting this.

> failed browser notifications

This is possibly a permissions issue? I'll have to dig a little deeper to see why it might be missing. But I should definitely be handling cases where notifications go wrong.

> I was wondering if we can use the /v1/payreq API call to decode the payment request.

I wanted to avoid this for latency reasons, my remote node can sometimes respond a little slow and I didn't want to add 500ms+ delays to payment prompts to decode the invoice. I'd rather contribute to `bolt11` and make that library more robust than require a network request.

bumi:
fyi: I was experimenting a bit with this and also moving parts into the background script.
My changes are in my fork - in case you want to follow: https://github.com/joule-labs/joule-extension/compare/autopay...bumi:autopay (includes loads of debug stuff for me to understand what is going on and get more familiar with the code)

bumi:
it seems the state of `crypto.isRequestingPassword` is always `true` [screenshot](https://i.imgur.com/IfyFcN0.png) - do you have an idea why that is the case? Shouldn't that be `false` once one payment is done and the checkbox is set to store the password for the session?

wbobeirne:
In that screenshot, password is still `null` so it would seem that the password hasn't yet been entered or fetched from the background script. The `hasSetPassword` indicates that they have ever set a password, not that they've entered it that session.

The password is only set again when the checkbox is set once it's needed. You can see [that logic over in the `crypto` saga](https://github.com/joule-labs/joule-extension/blob/develop/src/app/modules/crypto/sagas.ts#L30-L50).

bumi:
ok, thanks for the hint.
somehow it seems this is always the case. inspecting the state [here](https://github.com/joule-labs/joule-extension/compare/autopay...bumi:autopay#diff-3ced115a359b19b224273c76e9dd50c6R106) or [here](https://github.com/joule-labs/joule-extension/compare/autopay...bumi:autopay#diff-ca7a74490c1b54759e5468e52e89de38R56) gives me an empty password. 

Though I don't have to enter the password when I confirm the prompt. 
Is this related to the autopay changes? or some bug? I am currently stuck there any hints would be awesome :) - sadly I don't know the redux-saga concept yet.

wbobeirne:
Hey @bumi, sorry for leaving you hanging so long on this one. The password saving behavior is definitely a little tricky, and sagas are doubly tricky. The way it works is once you [enter your password with the save password checkbox set](https://github.com/joule-labs/joule-extension/blob/develop/src/app/modules/crypto/sagas.ts#L24-L28), it send a message to the background script ([sending code](https://github.com/joule-labs/joule-extension/blob/develop/src/app/utils/background.ts#L3-L9), [saving code](https://github.com/joule-labs/joule-extension/blob/develop/src/background_script/handlePassword.ts#L11-L14)) to hold onto it in memory.

Whenever a password request happens, we first check in with the background script if it has the password cached ([Request from background](https://github.com/joule-labs/joule-extension/blob/develop/src/app/modules/crypto/sagas.ts#L33-L40), [background sending back](https://github.com/joule-labs/joule-extension/blob/develop/src/background_script/handlePassword.ts#L21-L29)). Now, my understanding is that when that happens, it should fire the [`ENTER_PASSWORD` action that sets it all in state](https://github.com/joule-labs/joule-extension/blob/develop/src/app/modules/crypto/reducers.ts#L44-L49). I'm not entirely sure why you're not seeing what you're seeing in your logs, I'll do a little investigating of my own to see if I can reproduce it.

jamaljsr:
@wbobeirne I see you requested a review on this PR. I'm looking forward to playing with it. Is it ready? I should have time over the weekend.

wbobeirne:
Haha, oh man I totally forgot I had tagged you ages ago. Finally scratching out some free time to work on Joule again.

Planning on getting this in a stable state and trying to entice people to run it as a beta for me for some sats in exchange. Still has some work to go after the refactor in #244, but it should be pretty close.

Happily using Polar to do my testing with this right now.

jamaljsr:
No worries. I just saw a notification and wasn't sure if that was new or old. Feel free to let me know if you need any help with anything.

Once you go Polar, it's hard to go back to the terminal. Lol

bumi:
<3 I am also jumping back on this. let me know if I can help you with something.

wbobeirne:
Unfortunately in spite of moving the requests to the `background` context, we still get stopped with `ERR_SSL_PROTOCOL_ERROR` errors when trying to fetch to self signed certificate nodes. This is doubly bad because unlike the current workaround of going to the URL with the self signed cert and manually approving it, the `background` script will error out no matter what.

Unfortunately this type of automatic node communication may be dead in the water unless a native bridge is built (#106).

bumi:
hmm. does that mean current master also has that issue? 
I could try working on a native proxy for a showcase. But native extension can not be bundled with the extension, can they?
I think it should be prevented as official requirement as much as possible.

wbobeirne:
Yeah, that issue is currently present in `develop` (`master` is always the latest release, `develop` is the working branch.) I may revert those commits given these findings if I start working on something else, so that a new release doesn't break people's existing installations.

Native programs can't be bundled with extensions, they have to be downloaded and installed separately. This isn't an uncommon setup with some applications, like 1Password, where they have a version of the extension that requires a native application to be paired with it.

Not having native access has been a blocker for a ton of issues, such as switching to use the gRPC API, dealing with invalid TLS cers, being able to implement channel close, being able to handle `lightning:` URIs, and a ton of other things. While I've tried very hard to avoid going native, it definitely seems like it unlocks a ton of possibilities for Joule.

bumi:
fyi. this is the proxy app I've built to work around the invalid cert error: https://github.com/bumi/lnd-proxy
it runs a local proxy and proxies requests from localhost to the remote LND node.

wbobeirne:
Dude, this is very slick! I'd be happy to throw something in the readme after I look it over. Nice work.

bumi:
its a pretty early version and I have to learn some Golang for it :) And I am wondering about the UX and the usefulness of it. 
If the plan is to move towards a native bridge, then that could be a first step maybe.  

What is your plan with the allowance feature? As the communication through the background script should get reverted do you see any chance to implement the allowances? Could it maybe be optional?

curtcurt87:
webpack.config.js

curtcurt87 (commented):
webpack.config.js

PR #226, opened 2019-07-01
Add Lightning Loop functionality

### LINKED ISSUE #225: Add Support for Lightning Loop
### Integrate Lightning Loop in Joule UI

Lightning Loop is a non-custodial service offered by Lightning Labs to bridge on-chain and off-chain Bitcoin using submarine swaps. This repository is home to the Loop client and depends on the Lightning Network daemon lnd. All of lnd’s supported chain backends are fully supported when using the Loop client: Neutrino, Bitcoin Core, and btcd.
[Loop Github](https://github.com/lightninglabs/loop)

### Swagger Docs .yml for REST API
[swagger json](https://github.com/lightninglabs/loop/blob/master/looprpc/client.swagger.json)
image

callmekurisu:
Closes #225 

### Description

This PR attempts to address inbound liquidity issues by adding support for "Looping Out/In" to an off-chain address. The default address will be the lnd wallet address. The plan is to add a loop-http library similar to the lnd-http lib implementation. Will update as necessary pending changes to the lightning labs loop repo.

LND 0.7.0
Loop 2.2
* The branch below has the changes needed to build Loop In API changes
  you will also need to run ./gen_protos.sh in the looprpc directory
https://github.com/lightninglabs/loop/pull/74

### Steps to Test

1. Install Loopd 
2. View error handling for incorrect URL / Set correct Loop URL
3. View loop out terms (popover on advance settings buttons)
4. View Loop Out/In Quote
5. Attempt Loop Out/In to off-chain address (default LND wallet)
6. Use 'Loop Monitor' in terminal to check status
7. Default/Advanced User settings
8. Verify password before quote -> loop out/in

### Screenshot
image


wbobeirne:
Whoa, this is awesome. Great start! I'll keep an eye on this as you progress, but my main piece of feedback so far would be to make sure that loop stays an optional addition to Joule. A lot of people won't have loop running, and the eventual support of non-LND nodes means they won't have loop either. Right now the current implementation looks like it alters a lot of method signatures and validation checks to require a functioning loop server. Ideally it would only ask you to configure your loop server when you go to actually use the feature.

I'm really looking forward to seeing how this goes!

callmekurisu:
Thanks @wbobeirne! Good catch! I started to refactor so that the URL for Loop will be set optionally. Still need to add some snackbars so the user knows the URL was set successfully. Will try to finish adding the actual Loop out functionality tomorrow.

callmekurisu:
Getting close. Was able to attempt initiate a swap but it failed and Joule crashed too (T_T)

callmekurisu:
image

image

callmekurisu:
Initiated default loop out from Joule. Will add advanced features for the Loop Out form.

image

callmekurisu:
Added support for Loop In. Will create a separate PR for Loop Monitor. Need loop monitor to support synchronous requests with JSON array value.

image

callmekurisu:
Yo @wbobeirne check it out and let me know what you think!

wbobeirne:
Traveling for the next week but excited to see where it's at. Thanks for all of the hard work!

src/app/components/Loop/QuoteModal.tsx
```diff
@@ -0,0 +1,264 @@
...
+    ];
+    if (loopQuote === null) {
+      return null;
+    }
+    const isVisible = !!isOpen && !!(hasPassword || error);
+
+    let content;
+    if (loopQuote.miner_fee !== '') {
+      content = (
+        <>
```
wbobeirne: Unnecessary fragment, it's all in the `QuoteModal` div.

src/app/components/Loop/QuoteModal.tsx
```diff
@@ -0,0 +1,264 @@
...
+    }
+    return (
+      <Modal
+        title={`${type} Quote`}
+        visible={isVisible}
+        onCancel={onClose}
+        okButtonProps={{ style: { display: 'none' } }}
+        centered
+      >
+        <div className="QuoteModal">{content}</div>
```
wbobeirne: There's a div above with `QuoteModal` as a classname too, you probably don't want both of those.

src/app/components/Loop/QuoteModal.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+    this.props.getLoopIn(req);
+    setTimeout(() => {
+      message.info(`Attempting ${this.props.type}`, 2);
+    }, 1000);
+    setTimeout(() => {
+      this.props.onClose();
+    }, 3000);
+  };
+}
```
wbobeirne: I would consolidate the adv loop functions & buttons into the regular ones, and just form the `req` object based on if the `adv` prop is present or not.

src/app/components/Loop/QuoteModal.tsx
```diff
@@ -0,0 +1,264 @@
...
+  sct: string;
+  isOpen?: boolean;
+  type: string;
+  dest: string;
+  srf: string;
+  sf: string;
+  mf: string;
+  pre: string;
+  chan: string;
+  adv: boolean;
```
wbobeirne: Many of these props are hard to understand without context, I would keep them to their full names for better readability (even though it's somewhat more verbose.)

src/app/lib/loop-http/index.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  getLoopInTerms = () => {
+    return this.request<T.GetLoopTermsResponse>('GET', `/v1/loop/in/terms`);
+  };
+
+  getLoopOutQuote = (amt: string, conf: string) => {
+    return this.request<T.GetLoopQuoteResponse>(
+      'GET',
+      `/v1/loop/out/quote/${amt}?conf_target=${conf}`,
+      undefined,
```
wbobeirne: You should pass `{ conf_target: conf }` as the third arg here instead of doing interpolation and undefined.

src/app/utils/constants.ts
```diff
@@ -10,12 +10,15 @@ export enum NODE_TYPE {
...
   BTCPAY_SERVER = 'BTCPAY_SERVER',
+  LOOP = 'LIGHTNING LOOP',
 }
 
 export const DEFAULT_NODE_URLS = {
   [NODE_TYPE.LOCAL]: 'https://localhost:8080',
   [NODE_TYPE.LIGHTNING_APP]: 'https://localhost:8086',
   [NODE_TYPE.ZAP_DESKTOP]: 'https://localhost:8180',
+  // Need to update to https after testing
+  [NODE_TYPE.LOOP]: 'http://localhost:8081',
```
wbobeirne: I'm pretty sure Loop will normally run over HTTP since there're no credentials

src/app/components/Loop/index.tsx
```diff
@@ -0,0 +1,425 @@
...
+      <Menu.Item
+        key={c.channel_point}
+        onClick={() => this.handleSetChannelId(c.channel_point, openChannelsLoopIn)}
+      >
+        {`${c.node.alias} => ${c.local_balance} sats available`}
+      </Menu.Item>
+    ));
+
+    const loopInMenu = <Menu>{loopInItems}</Menu>;
+    const loopOutMenu = <Menu>{loopOutItems}</Menu>;
```
wbobeirne: These behave really poorly as `<Menu>`s (especially if you have a lot of channels.) I think they should be `<Select>` lists instead.

src/app/components/Loop/index.tsx
```diff
@@ -0,0 +1,425 @@
...
+            )}
+            {url !== null && loopType === 'Loop In' && (
+              <Form.Item>
+                <Dropdown overlay={loopInMenu}>
+                  <a className="ant-dropdown-link" href="#">
+                    Select Channel
+                  </a>
+                </Dropdown>
+              </Form.Item>
+            )}
```
wbobeirne: These two behave identically, they probably shouldn't be seperate conditionals but should instead be:
```
<Dropdown overlay={loopMenu}>
```
where loopMenu is a variable that's either loopInMenu or loopOutMenu

src/app/components/Loop/index.tsx
```diff
@@ -0,0 +1,425 @@
...
+            {advanced === true && loopType === 'Loop Out' && (
+              <Form.Item label="">
+                <Input
+                  size="small"
+                  onChange={this.handleChangeSwapRoutingFee}
+                  placeholder="swap routing fee"
+                  autoFocus
+                />
+              </Form.Item>
+            )}
```
wbobeirne: Same as above, these are mostly the same

src/app/components/Loop/index.tsx
```diff
@@ -0,0 +1,425 @@
...
+    this.setState({ minerFee: ev.currentTarget.value });
+  };
+
+  private handleChangePrepayAmt = (ev: React.ChangeEvent<HTMLInputElement>) => {
+    this.setState({ prepayAmt: ev.currentTarget.value });
+  };
+
+  private handleChangeConfTarget = (ev: React.ChangeEvent<HTMLInputElement>) => {
+    this.setState({ conf: ev.currentTarget.value });
+  };
```
wbobeirne: All of these handles could be a single `handleChangeField` method that sets it based on the input's `name` property.

src/app/components/Loop/index.tsx
```diff
@@ -0,0 +1,425 @@
...
+  destination: string;
+  swapRoutingFee: string;
+  swapFee: string;
+  minerFee: string;
+  prepayAmt: string;
+  channel: string;
+  conf: string;
+  htlc: boolean;
+  quoteModalIsOpen: boolean;
+  loopType: string;
```
wbobeirne: `loopType` should be an `enum` to avoid making mistakes with the magic string.

src/app/components/Loop/index.tsx
```diff
@@ -0,0 +1,425 @@
...
+              <Form.Item>
+                <Input
+                  size="small"
+                  onChange={this.handleChangeConfTarget}
+                  placeholder="sweep confirmation target"
+                  autoFocus
+                />
+              </Form.Item>
+            )}
+            {advanced === true && loopType === 'Loop In' && (
```
wbobeirne: You can replace all `advanced === true` checks with just `advanced &&...`

src/app/components/Loop/index.tsx
```diff
@@ -0,0 +1,425 @@
...
+              title={`
+                Base fee: ${loopTerms.swap_fee_base} sats |
+                Fee rate: ${loopTerms.swap_fee_rate} sats |
+                Prepay amt: ${
+                  loopTerms.prepay_amt === undefined ? '1337' : loopTerms.prepay_amt
+                }
+                Min Swap amt: ${loopTerms.min_swap_amount} sats |
+                Max Swap amt: ${loopTerms.max_swap_amount} sats |
+                CLTV delta: ${loopTerms.cltv_delta}
+              `}
```
wbobeirne: This tooltip displays way too much data to work well as a tooltip. It should either be persistently rendered inline, or toggle / collapsable.

src/app/modules/loop/reducers.ts
```diff
@@ -0,0 +1,211 @@
+import types, { LoopTermsPayload, LoopQuotePayload, LoopPayload } from './types';
+import LoopHttpClient from 'lib/loop-http';
+
+export interface LoopState {
+  loopTerms: null | LoopTermsPayload;
+  loopQuote: null | LoopQuotePayload;
```
wbobeirne: Are quote and terms ever actually null? Looks like they're initialized with empty fields.

src/app/modules/loop/reducers.ts
```diff
@@ -0,0 +1,211 @@
...
+    case types.SET_LOOP_IN_FAILURE:
+      return {
+        ...state,
+        error: action.payload,
+        lib: null,
+      };
+    case types.GET_LOOP_IN_TERMS:
+      return {
+        ...state,
+      };
```
wbobeirne: If an action does nothing to state, it doesn't need to be a case in the reducer.

src/app/modules/loop/reducers.ts
```diff
@@ -0,0 +1,211 @@
...
+    case types.GET_LOOP_IN_QUOTE_FAILURE:
+      return {
+        ...state,
+        error: action.payload,
+      };
+    case types.GET_LOOP_IN:
+    case types.GET_LOOP_OUT_TERMS:
+      return {
+        ...state,
+      };
```
wbobeirne: ^ Same as above

src/app/modules/loop/reducers.ts
```diff
@@ -0,0 +1,211 @@
...
+    case types.GET_LOOP_OUT_QUOTE_FAILURE:
+      return {
+        ...state,
+        error: action.payload,
+      };
+    case types.GET_LOOP_OUT:
+      return {
+        ...state,
+      };
```
wbobeirne: ^ Same as above

src/app/modules/loop/reducers.ts
```diff
@@ -0,0 +1,211 @@
...
+        loop: action.payload,
+      };
+    case types.GET_LOOP_IN_FAILURE:
+      return {
+        ...state,
+        error: action.payload,
+      };
+      return {
+        ...state,
+      };
```
wbobeirne: This return will never execute

src/app/modules/loop/sagas.ts
```diff
@@ -0,0 +1,149 @@
...
+
+// Setup Loop URL and Loop Terms
+export function* handleSetLoopOut(
+  action: ReturnType<typeof actions.setLoop>,
+): SagaIterator {
+  const url = action.payload;
+  let loopTermsPayload;
+  try {
+    const client = new LoopHttpClient(url);
+    loopTermsPayload = yield call(client.getLoopInTerms);
```
wbobeirne: Should use `getLoopOutTerms`

src/app/lib/loop-http/index.ts
```diff
@@ -0,0 +1,126 @@
...
+    );
+  };
+
+  getLoopIn = (args: T.GetLoopInArguments) => {
+    return this.request<T.GetLoopResponse, T.GetLoopInArguments>(
+      'POST',
+      '/v1/loop/in',
+      args,
+    );
+  };
```
wbobeirne: These are POST requests, not GETs, so I think they should be renamed to `postLoop...`, `makeLoop...` or something like that. Ideally this nomenclature could be renamed everywhere, as it's a little confusing to call a function that starts with `get` that causes a payment to happen.

callmekurisu:
Thanks for the review! Submitted all the recommended changes on one big commit. If you're able to take it from here and work your magic it would be much appreciated. I'm learning a lot! Hope those API changes make it to the next loop release. I will continue learning Go and gRPC so I can work more on loop and lnd. Nice working with you!

src/app/components/Loop/InputLoopAddress.tsx
```diff
@@ -0,0 +1,94 @@
...
+            value={url}
+            onChange={this.handleChange}
+            placeholder="http://localhost:8081"
+            autoFocus
+          />
+        </Form.Item>
+
+        <Button type="primary" size="large" htmlType="submit" disabled={!url} block>
+          {`Set ${type} URL`}
+        </Button>
```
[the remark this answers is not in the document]
callmekurisu: added loading state

src/app/components/Loop/QuoteModal.tsx
callmekurisu: removed fragment

src/app/components/Loop/QuoteModal.tsx
callmekurisu: changed classname

src/app/components/Loop/QuoteModal.tsx
callmekurisu: consolidated advanced loops with ternary

src/app/components/Loop/QuoteModal.tsx
callmekurisu: I just got lazy there! My bad. Made it more verbose!

src/app/lib/loop-http/index.ts
callmekurisu: updated third arg

src/app/utils/constants.ts
callmekurisu: Ah shoot, yes that's correct. Removed that comment

src/app/components/Loop/index.tsx
callmekurisu: The Select is looking much better for sure. It was super hacky before. Only problem I was seeing now is that console complains that I should use Select.Option, but then typescript complains that Select.Options has no onClick prop. So I left as Menu.Item

src/app/components/Loop/index.tsx
callmekurisu: updated with ternary

src/app/components/Loop/index.tsx
callmekurisu: removed

src/app/components/Loop/index.tsx
callmekurisu: Wow, did not know this was possible. Very cool.

src/app/components/Loop/index.tsx
callmekurisu: added loop type to constants.ts

src/app/components/Loop/index.tsx
callmekurisu: replaced all advance === true to advanced &&

src/app/components/Loop/index.tsx
callmekurisu: Replaced tooltip with collapse

src/app/modules/loop/reducers.ts
callmekurisu: Shoots, that's right. Not sure what I was thinking. Removed null

src/app/modules/loop/reducers.ts
callmekurisu: Cleaned up entire reducer

src/app/modules/loop/reducers.ts
callmekurisu: Cleaned up reducer

src/app/modules/loop/reducers.ts
callmekurisu: Cleaned up reducer

src/app/lib/loop-http/index.ts
callmekurisu: removed "get" from the post requests.

src/app/components/Loop/InputLoopAddress.tsx
```diff
@@ -0,0 +1,94 @@
...
+        origins: [urlWithoutPort(url)],
+      })
+      .then(accepted => {
+        if (!accepted) {
+          message.warn('Permission denied, connection may fail');
+        }
+        this.setState({ submittedUrl: url });
+        loop.type === 'Loop Out'
+          ? loop.setLoop(this.state.url)
+          : loop.setLoopIn(this.state.url);
```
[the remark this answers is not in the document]
callmekurisu: they shouldn't be different urls. I went back and got rid of "set loop in". loop is set and terms are retrieved asynchronously now.

src/app/modules/loop/reducers.ts
callmekurisu: cleaned up reducer

wbobeirne:
Great work. I'm getting started on a bit of refactoring, will have you review when I'm done. Looking forward to getting this live 🔥

I'll open a new PR for that, so I'm going to close this one for the time being just to make sure no new work or reviews end up on this branch that I have to merge back into the one I'm working on.

PR #253, opened 2020-05-24
Store last payment attempt in the allowance config

bumi:
### Description

So far the `lastPaymentAttempt` is not presisted yet and thus the cooldown period for an allowance could not be calculated. 
This change stores the `lastPaymentAttempt` timestamp as part of the allowance config. The default value is 0 as before.

### DIFF
--- a/src/content_script/respondWithoutPrompt.ts
+++ b/src/content_script/respondWithoutPrompt.ts
@@ -12,8 +12,6 @@ import {
 } from '../util/messages';
 import { createNotification, updateNotification } from './notifications';
 
-let lastPaymentAttempt = 0;
-
 export default async function respondWithoutPrompt(
   msg: AnyPromptMessage,
 ): Promise<boolean> {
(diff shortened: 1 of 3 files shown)


src/content_script/respondWithoutPrompt.ts
```diff
@@ -63,7 +63,7 @@ async function handleAutoPayment(msg: PaymentPromptMessage) {
   }
 
   // Don't allow payments to happen too fast
-  const last = lastPaymentAttempt;
+  const last = allowance.lastPaymentAttempt || lastPaymentAttempt;
```
wbobeirne, line 66: This should probably be `Math.max`, we probably want to defer to whichever is most recent.

bumi: actually I think I can remove the `|| lastPaymentAttempt` because `allowance.lastPaymentAttempt` should be `0` for new allowances: https://github.com/joule-labs/joule-extension/pull/253/files#diff-34fdfa4008f4ad6d136ac8f4335e804aR259
and then we can rely on `allowance.lastPaymentAttempt` being updated.

bumi: changed it to rely on the data persisted in the allowance config.

wbobeirne:
Awesome, thanks for revisiting!

PR #254, opened 2020-05-24
run yarn upgrade-interactive

bumi:
I ran `yarn upgrade-interactive` as a start to update the dependencies. 

@svgr/webpack needed to be updated to solve SVG errors after updating
babel/core:
```
    ERROR in ./src/app/static/images/zap.svg
    Module build failed (from ./node_modules/@svgr/webpack/lib/index.js):
    Error: Cannot find module '@babel/plugin-transform-react-constant-elements' from 'lightning/joule-extension'
```

I somehow could not catch the following error. But I also don't have that one in the `autopay` branch: 
image
this is after an authorize call. 

`messageWithOrigin` is:  `{application: "Joule", prompt: true, type: "AUTHORIZE", args: undefined, origin: {…}}`

`response` seems to be `null` here: https://github.com/joule-labs/joule-extension/blob/develop/src/content_script/index.ts#L29

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "@babel/core": "^7.0.1",
     "@babel/preset-env": "7.6.0",
     "@babel/preset-react": "7.0.0",
-    "@svgr/webpack": "3.1.0",
+    "@svgr/webpack": "5.4.0",
     "antd": "^3.26.12",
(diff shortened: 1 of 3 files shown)


bumi:
closing this in favor of a new one with more fixes and updates.

Issue #255, opened 2020-05-24
Setup depfu to keep dependencies up to date


bumi:
Currently quite some dependencies need to be updated. I've tried to start doing that manually but it is quite a hassle. Maybe we can setup [depfu](https://depfu.com/) to keep them up to date.

Depfu creates PRs for each dependency which makes it easy to review them step by step. Depfu can also create PRs for out-of-spec dependencies to keep those updated even outside of the specified constraints.

To test how it works for joule I've added it to [my fork](https://github.com/bumi/joule-extension/pulls)

wbobeirne:
Hey @bumi thanks for this and the PR. My attitude about dependencies is to try and only upgrade them for security vulnerabilities (Which I had Snyk setup to monitor, but it looks like I've fallen behind on a few) or if there's a particular bugfix or feature from a new version I find I need.

Were there any particular packages you felt would be worth upgrading for new features (e.g. antd) or were you hoping to just keep all of them up to date?

bumi:
ah I see. I was just trying to keep them up to date as I feel it gets harder to update the more out of date they are. And at some point older versions also don't get security updates anymore.

wbobeirne:
We're all up to date now and I'm gonna keep dependabot going for important updates.

PR #257, opened 2020-06-04
Update dependencies

bumi:
I was running depfu against may fork to update all the dependencies and I will try to keep that [branch](https://github.com/bumi/joule-extension/tree/dependency-updates) up to date with latest releases. 
Opening this PR in a reference to #255  - maybe the branch is helpful for some updates at some point
Feel free to close the PR. 

### Required code changes to update: 

* no `>` in the HTML: https://github.com/joule-labs/joule-extension/commit/5132b6166102cc678b844c9addd243e8788837bf
* new options API for less-loader: https://github.com/joule-labs/joule-extension/commit/5e653cfc31b2f8eda0af223a72a04f116861aa87
* linting errors after tslint update: https://github.com/joule-labs/joule-extension/commit/3c62ff7f6b7aaaa104f5411340f7ca37a1651343
* new prettier style: https://github.com/joule-labs/joule-extension/commit/6af123213e73aecf96125618041631432e823136
* `Object.assign` instead of spread operator: https://github.com/joule-labs/joule-extension/commit/2d1fe0437d84c9e5c7168c1c267a049ff75c19f9

### Current issues: 

* tslint is deprecated and should be replaced with eslint & Co. https://github.com/palantir/tslint/issues/4534
*  ~~tsc issues: https://travis-ci.org/github/joule-labs/joule-extension/builds/694760264#L225 - are those real errors? (mainly in: https://github.com/joule-labs/joule-extension/blob/develop/src/lnd/http/index.ts)~~
* major changes in [ant design](https://ant.design) - [migration guide v3 -> v4](https://ant.design/docs/react/migration-v4)
* window sizing / spacing error with latest [css-loader](https://github.com/webpack-contrib/css-loader) (see: https://github.com/bumi/joule-extension/pull/72#issuecomment-639005016)

### Links regarding the TypeScript errors: 

* ~~https://github.com/microsoft/TypeScript/issues/30188~~
* ~~https://github.com/microsoft/TypeScript/issues/34933~~
* ~~https://stackoverflow.com/questions/57798016/how-to-ignore-type-instantiation-is-excessively-deep-and-possibly-infinite-ts~~
* ~~https://github.com/microsoft/TypeScript/issues/38535~~

### DIFF
--- a/src/app/components/SendForm/ChainSend.tsx
+++ b/src/app/components/SendForm/ChainSend.tsx
@@ -202,7 +202,7 @@ class ChainSend extends React.Component<Props, State> {
 }
 
 export default connect<StateProps, DispatchProps, OwnProps, AppState>(
-  (state) => ({
+  state => ({
     account: state.account.account,
     sendOnChainReceipt: state.payment.sendOnChainReceipt,
     isSending: state.payment.isSending,
--- a/src/app/components/SendForm/ChainSend.tsx
+++ b/src/app/components/SendForm/ChainSend.tsx
@@ -94,9 +94,7 @@ class ChainSend extends React.Component<Props, State> {
       <>
         <Icon type="exclamation-circle" /> Dust warning: This amount may not be spendable
       </>
-    ) : (
-      undefined
-    );
+    ) : undefined;
 
     return (
       <Form className="ChainSend" layout="vertical" onSubmit={this.handleSubmit}>
@@ -204,7 +202,7 @@ class ChainSend extends React.Component<Props, State> {
 }
 
 export default connect<StateProps, DispatchProps, OwnProps, AppState>(
-  state => ({
+  (state) => ({
     account: state.account.account,
     sendOnChainReceipt: state.payment.sendOnChainReceipt,
     isSending: state.payment.isSending,
--- a/.prettierrc
+++ b/.prettierrc
@@ -6,5 +6,6 @@
   "tabWidth": 2,
   "trailingComma": "all",
   "bracketSpacing": true,
-  "jsxBracketSameLine": false
+  "jsxBracketSameLine": false,
+  "arrowParens": "avoid"
 }
--- a/package.json
+++ b/package.json
@@ -117,7 +117,7 @@
     "@types/redux-logger": "^3.0.8",
     "@types/remote-redux-devtools": "0.5.4",
     "@types/semver": "7.3.1",
-    "@types/webpack": "^4.41.21",
+    "@types/webpack": "^4.41.22",
     "@types/zxcvbn": "4.4.0",
     "cross-env": "7.0.2",
     "pretty-quick": "3.0.2",
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "crypto-js": "4.0.0",
     "css-loader": "^1.0.0",
     "file-loader": "^6.1.0",
-    "html-webpack-plugin": "4.3.0",
+    "html-webpack-plugin": "4.5.0",
     "husky": "4.2.5",
     "jdenticon": "3.0.1",
     "less": "^3.12.2",
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "jdenticon": "2.2.0",
     "less": "^3.12.2",
     "less-loader": "^6.2.0",
-    "mini-css-extract-plugin": "^0.9.0",
+    "mini-css-extract-plugin": "^0.11.2",
     "moment": "^2.27.0",
     "node-sass": "^4.14.1",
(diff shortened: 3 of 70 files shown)


bumi:
The TypeScript errors got fixed mainly by another TypeScript update. 
And I had to replace the use if the spread operator at a few spots with `Object.assign`: https://github.com/joule-labs/joule-extension/pull/257/commits/2d1fe0437d84c9e5c7168c1c267a049ff75c19f9
With those updates CI is now green.

wbobeirne:
Awesome, I'm probably going to gear up to focus on build reproducibility soon, so I'll be revisiting this PR in the near future. Unfortunate about the spread bugs, but thanks for including the details in your commit!

bumi:
Turns out I missed an error: 

image
in: https://github.com/joule-labs/joule-extension/blob/4f7a333a508728364d2ddd7155b76cde926f3ad7/src/content_script/index.ts#L34

the message that is sent:
```js
messageWithOrigin = { 
  application: "Joule",
  args: undefined,
  origin: {domain: "http://localhost:4567", name: "402", icon: "http://localhost:4567/favicon.ico"},
  prompt: true,
  type: "AUTHORIZE",
}
```

any idea what could cause that error? 
According to the [sendMessage docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/sendMessage) the promise canbe  fulfilled with no arguments if the receiver does not send a response.

bumi:
that error is also present in the develop branch and it seems it was introduced in https://github.com/joule-labs/joule-extension/commit/85bac5ccdb62632ea4940511fdb69ffde00931c4
but is not present it the autopay branch. 
so I guess it's ok.

wbobeirne:
Ah thanks for the catch. Based on my findings in https://github.com/joule-labs/joule-extension/pull/222#issuecomment-593745595, I may end up reverting that commit, because the background script behavior should it be released would break the extension for the majority of users.

bumi:
yep, which is a pity. I think the allowance feature is pretty cool.
Currently I experiment with it using a local [caddy server](caddyserver.com) as proxy to my node.

bumi:
@wbobeirne what are your thoughts on how to best continue with the different topics? 
What do you think about updating the dependencies (this PR)
will you revert the background script behavior? (which would be a pity)

I am afraid that merging will get more complicated

wbobeirne:
Hey @bumi, thanks for keeping on this. Let's get your PR in ASAP so that it doesn't run into conflicts. From there I'll revert the background behavior (but keep the git history to try to address it in the future) so that we don't bust anything for anyone. From there, I'd like to get a consistent build process (maybe with a docker container) and resubmit to the Firefox extension store, and just do a minor release for that.

src/app/components/SendForm/ChainSend.tsx
```diff
@@ -204,7 +202,7 @@ class ChainSend extends React.Component<Props, State> {
 }
 
 export default connect<StateProps, DispatchProps, OwnProps, AppState>(
-  state => ({
+  (state) => ({
```
wbobeirne, line 205: Can we enable https://prettier.io/docs/en/options.html#arrow-function-parentheses to reduce the size of the diff with the prettier upgrade?

wbobeirne (commented):
Just the one request, then let's get this thing merged in.

src/app/components/SendForm/ChainSend.tsx
bumi: :+1: 
done in https://github.com/joule-labs/joule-extension/pull/257/commits/2998810ead18c51e105a306798e1c02e25453e56 
configuration changed in https://github.com/joule-labs/joule-extension/pull/257/commits/f39e9fd5f2fbf730d221a1b575a1d3ac9a4210b3

bumi:
sounds good! some of the dependencies got some more updates, I'll pull those in tomorrow.

bumi:
ok, done. now only these issues are left: 

* css-loader issue (https://github.com/bumi/joule-extension/pull/72#issuecomment-639005016) - Do you have any idea what the problem could be here? I have no idea how to approach this one
* switch to eslint
* ant design upgrade to v4

Those should be done in separate pull requests.

wbobeirne (approved):
Awesome work! Sorry for the delay.

* For the CSS Loader issue, I'm willing to bet it has something to do with style load order. I'll try to take a look there.
* Switching to ESLint sounds good, unfortunately in my experience with other projects it's got painfully slower, but I'd prefer that over falling behind. I may have to rework the rules a bit though.
* Ant design upgrade is going to suck, since they actually have changed the way things look a good bit. I'll have to stew on this one.

PR #266, opened 2020-11-05
Add GitHub workflow to run build:prod

bumi:
This GitHub workflow runs a production build on every push to the develop branch (e.g. whenever a pull request is merged) and whenever a version tag is created.
The build artifacts are stored and made available for download. 

You can see the workflow in action here: https://github.com/bumi/joule-extension/actions/runs/346573887

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,6 +5,8 @@ on:
   push:
     branches:
       - develop
+    tags:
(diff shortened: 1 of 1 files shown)


wbobeirne (approved):
Awesome, I'll probably expand this to also do releases as well so that the assets end up over there. Thanks!

Issue #260, opened 2020-07-23
Joule no longer available as firefox extension


ArbenKarb:
### Description of the Feature or Idea

[Your description goes here]
![image](https://user-images.githubusercontent.com/31412198/88321420-6fb47880-cd1f-11ea-82eb-5c14fe337dfd.png)

### Existing Example(s) of Feature

[If another product or app is doing this, please link to it]

wbobeirne:
Argh, looks like it got removed because they weren't able to reproduce the builds locally from the source provided (different checksums.) In the meantime, it can be manually installed from the release assets: https://github.com/joule-labs/joule-extension/releases/tag/v0.5.2. I'll work on getting it restored in the store though.

bumi:
bump, is there anything we can do to get it back up?

githorray:
love bump

bumi:
The main problem is that zip files contain the file modification timestamps and thus the build never fully reproducible. 
The problem of timestamps and reproducible builds is described [here](https://reproducible-builds.org/docs/timestamps/)

To work around this issue we can use the strip-nondeterminism tool ([debian package](https://packages.debian.org/sid/strip-nondeterminism), [git repo](https://salsa.debian.org/reproducible-builds/strip-nondeterminism)) which knows how to normalize typical files like zip files. (by setting a fixed modification timestamp on each file)

Or we update the files manually before zipping as for example described [here](https://medium.com/@pat_wilson/building-deterministic-zip-files-with-built-in-commands-741275116a19). This would need to happen before creating the zip. 
(It seems there is also a not very much used [npm package](https://www.npmjs.com/package/deterministic-zip) that claims to make deterministic zip files)

For now I used the strip-nondeterminism tool, just because using that I did not have to change the build process and could simply update the zip file.

I have created a Dockerfile to run the the build process inside a container and to create the zip package. PR: #268 
This should solve this problem and allow mozilla build and verify the add-on file.


* [Notes on Source Code submission](https://extensionworkshop.com/documentation/publish/source-code-submission) (Checklist and notes on how to provide build instructions)
(It seems that Mozilla also recommends to NOT minify the code and we probably do not need sourcemaps or similar in production. Do you think of anything that we should change in the production build process?)

bumi:
Here is another option which I actually prefer: https://github.com/joule-labs/joule-extension/pull/269
It changes the build process slightly to not create the zip with JS in the webpack build process but using a small bash script.
[...]

PR #268, opened 2020-11-05
Add Dockerfile to create production builds of Joule

bumi:
Started a Dockerfile that can be used to create production builds of joule. 
This should make it easier for people to create and verify production builds.

I am not sure what the best approach for this is using docker. 
Currently the docker image needs to be created: 

    $ docker build -t joule .

And then the production .zip can be copied to the host machine:

    $ docker run --rm -v "$(pwd):/out" joule sh -c "cp /app/dist-prod/joule*.zip /out"

joule-vx.x.x.zip will then be in the project root folder.

Alternatively run the production build from the current working directory:

    $ docker run --rm -v "$(pwd):/build" joule sh -c "cd /build && yarn && yarn build"

All build dependencies run in Docker but will create the build `dist-prod` on the host.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,11 +8,16 @@
 # $ docker run --rm -v "$(pwd):/out" joule sh -c "cp /app/dist-prod/joule*.zip /out"
 # ls ./out
 #
-FROM node:lts-alpine
+FROM node:lts
+run apt-get update
+RUN apt-get install strip-nondeterminism --yes
+
 WORKDIR /app
 COPY . /app
 RUN rm -rf /app/dist-prod # remove potentially old folder
 RUN yarn
 RUN yarn build
+RUN strip-nondeterminism /app/dist-prod/joule-*.zip
+RUN sha256sum -b /app/dist-prod/joule-*.zip
 #COPY /app/dist-prod/joule-v0.5.2.zip /out
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+# This Dockerfile can be used to build a production release of joule:
+#
+# Usage:
+# 1. Build the docker image
+# 2. Get the production distribution .zip file from the image
+
+# $ docker build -t joule .
+# $ docker run --rm -v "$(pwd):/out" joule sh -c "cp /app/dist-prod/joule*.zip /out"
+# ls ./out
+#
+FROM node:lts-alpine
+WORKDIR /app
+COPY . /app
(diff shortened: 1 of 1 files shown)


bumi:
Actually I think this does not make sense and this is way simpler and we do not need a Dockerfile at all: 

```
$ docker run --rm -v "$(pwd):/app" node:lts sh -c "apt-get update && apt-get install zip --yes &&cd /app && yarn install && yarn build && yarn package:prod
```
this mounts the current directory in a node container, installs zip, builds the production dist and zips the package using the script in #269

This does not require anything on the host machine (except docker) and creates a verifiable build. (to solve #260)

wbobeirne:
That's pretty slick. I think I still like the Dockerfile approach, since we can split it out into multiple lines and it's a bit easier to parse that way. It would still have the same dependency requirements (just docker) but be a bit easier to maintain. It's also nice for people who want to develop in a Docker container, and don't want to install node et al on their host machine.

Thanks for getting started on this, I'll try to play with this and #269 a bit this week and confirm the builds are nice'n'deterministic.

bumi:
OK. great! 

I'd combine those commands in a build script and the docker command would then be something like: 
`docker run --rm ----volume="$(pwd):/app" --workdir="/app" node:lts /app/scripts/build`
It depends a bit on how we want to build the deterministic zip (#269 or something else) 

If we add the Dockerfile to allow people developing with it I think we should somehow make sure it can easily be used to run dev commands. 

But in the end the only dependency is node and `yarn install` needs to run. This can also easily be done from within the default node container: 
`$ docker run --rm -v "$(pwd):/app" --workdir="/app" -t -i node:lts bash`  - this gives you a shell to run the yarn commands.

bumi:
This actually also looks quite nice for our usecase: https://earthly.dev/

bumi:
@wbobeirne any thoughts on this? I think it would be nice to get joule back into the firefox store.

bumi:
@wbobeirne sorry for bothering, but do you have any thoughts on this?
This should work and if we consider joule useful and want to progress I think it would be nice to get it back in the browser add-on stores.

wbobeirne:
I've carried this work on in #295, unfortunately despite the name, `strip-nondeterminism` does not do as advertised. I've tried also passing it `--timestamp` but it doesn't seem to rewrite the file last modified times. The zip gets a new hash every time.

Going to close this PR and continue iterating in the new PR.

PR #269, opened 2020-11-07
Create deterministic and verifiable zip package

bumi:
This is another option to create reproducible zip packages that can be verified. 
To do this we no longer create the zip in the webpack build process but using a small bash script.

The script sets a fixed file timestamp to the `dist-prod` files and creates the zip file. 

related: #260  #268

### DIFF
--- a/create-package.sh
+++ b/create-package.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+# Extract version from package.json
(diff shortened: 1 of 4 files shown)


wbobeirne:
Ended up generating a hash manifest of zip contents instead of trying to make the zip reproducible in #295, so closing this effort. Thanks for all the contributions for reproducability!

PR #271, opened 2020-12-15
Dependency updates

bumi:
This is another round of dependency updates from depfu. 

The main update is react 17. 
Everything seemed fine for me, but some more tests are needed.


wbobeirne (approved):
Righteous, thanks!

PR #276, opened 2021-05-02
Reflect invoice expiry in payment popups #228

### LINKED ISSUE #228: Confirm and Reject payment buttons should reflect Invoice expiry
### Description of the Issue

When an invoice has expired, Joule should not allow the 'Confirm' payment button to be clicked. It should be disabled, gray, and/or show 'Expired'.

Additionally, the Reject button is no longer appropriate after expiry. It should simply be replaced with a 'Close' button.

### Steps to Reproduce

Create an invoice that expires, and then leave the payment dialogue box open until after expiry

### Description of Your Setup

*Browser & version*: Brave
*Operating system*: Linux
*Node software & version*: LND 0.7.1-beta

[Please provide these if you can!]

szche:
Closes #228 

### Description

As descirbed in issue #228, popup payments should check wheter the invoice has expired or not. This merge borrows some of the code from "Send Payment" tab. If the invoice is in fact expired, the "Confirm" button gets grayed out and disabled.

### Steps to Test
Try to pay an invoice that is still valid and the one that has already expired

### Screenshots
Quick before and after view of a sample expired invoice
image

### DIFF
--- a/src/app/prompts/payment.tsx
+++ b/src/app/prompts/payment.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import moment from 'moment';
 import { connect } from 'react-redux';
-import { Tabs, Input, Select } from 'antd';
+import { Tabs, Input, Select, Alert } from 'antd';
 import { SendPaymentResponse } from 'webln';
 import PromptTemplate from 'components/PromptTemplate';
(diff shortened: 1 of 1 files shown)


wbobeirne (approved):
Great improvement, thanks!

PR #295, opened 2021-12-16
Docker-based deterministic builds

wbobeirne:
Continuation of #268

### Description

* Split linting and building into separate workflows
  * Linting will run on all PRs
  * Builds will only run on develop / master merges and tags
* Build in CI using docker for deterministic builds
* Add handy `yarn build:docker` command to make it easy

### Steps to Test

1. Watch CI, confirm it passes
2. Run CI multiple times, confirm the same hash is generated

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,5 @@
+# dependencies
+/node_modules
+
+# production
+/dist-*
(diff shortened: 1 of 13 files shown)


wbobeirne:
This is looking pretty Sisyphean, I think I might be better off with an sha256 / md5 of all the files _in_ the zip instead.

wbobeirne:
I can verify the sha256sum manifest from the docker build locally matches the one from GitHub actions 🙌

PR #305, opened 2022-02-21
Add BRL fiat currency

### LINKED ISSUE #304: BRL Fiat pair
Add `BRL` to the fiat options.

### Existing Example(s) of Feature

There are `USD, EUR, GBP, JPY, RUB` already, why not add another one? 🤷🏾‍♂️

johnnyasantoss:
Closes #304 

### Description

[Description of your changes go here. Please provide both casual and technical explanations.]

### Steps to Test

1. Open joule
2. Go to the settings page
3. Change the fiat pair to BRL
4. Check if it worked on the main page

### Screenshots

[Only if applicable]

### DIFF
--- a/src/app/utils/constants.ts
+++ b/src/app/utils/constants.ts
@@ -202,6 +202,7 @@ export enum Fiat {
   GBP = 'GBP',
   JPY = 'JPY',
   RUB = 'RUB',
+  BRL = 'BRL',
 }
 
 export const fiatSymbols: { [key in Fiat]: string } = {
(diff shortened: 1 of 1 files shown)


wbobeirne (approved):
Awesome, thanks for the contribution! Looks good to me. The fiat pairs will probably need a little reworking before we add too many more after this though, currently each one is a separate network request that we do all at once. Very silly, we should only be requesting the pair a user has configured. But one more won't hurt for now!

johnnyasantoss:
> Awesome, thanks for the contribution! Looks good to me. The fiat pairs will probably need a little reworking before we add too many more after this though, currently each one is a separate network request that we do all at once. Very silly, we should only be requesting the pair a user has configured. But one more won't hurt for now!

Yeah I noticed while following the path to make sure what needed to be changed to add a new pair. 
I think that requesting only the one selected would already be good enough. I could make that change but I still don't understand how sagas work exactly.
I need to study that a bit more.

Anyways, great extension, thanks a lot!

